PR #4, opened 2018-08-24
U/lynnej/add eups sal

rhiannonlynne:
Add eups & lsstsw build capability into ts_xml (and thus ts_sal).


rhiannonlynne:
Merging branch u/lynnej/add_eups_sal into scheduler/develop

PR #22, opened 2018-10-23
TSS-3197: ScriptLoader xml updates

r-owen:


r-owen:
@tribeiro good catch. Fixed.

PR #30, opened 2018-10-26
Feature/andrewheyer/final scheduler update

andrewbheyer:


sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -82,114 +166,114 @@
...
-    <EFDB_Topic>MTScheduler_logevent_settingsApplied</EFDB_Topic>
+    <EFDB_Topic>Scheduler_logevent_settingsApplied</EFDB_Topic>
     <Alias>settingsApplied</Alias>
     <Explanation>http://sal.lsst.org</Explanation>
     <item>
         <EFDB_Name>version</EFDB_Name>
         <Description></Description>
         <IDL_Type>string</IDL_Type>
-        <IDL_Size>256</IDL_Size>
+        <IDL_Size>1</IDL_Size>
```
tribeiro: Is this right? I imagine the size here should be 256 no? There are other places with the same issue.

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -82,114 +166,114 @@
...
     </item>
 </SALEvent>
 <SALEvent>
-    <Subsystem>MTScheduler</Subsystem>
-    <Version>3.7</Version>
+    <Subsystem>Scheduler</Subsystem>
+    <Version>3.8.13</Version>
     <Author></Author>
-    <EFDB_Topic>MTScheduler_logevent_validSettings</EFDB_Topic>
+    <EFDB_Topic>Scheduler_logevent_validSettings</EFDB_Topic>
```
tribeiro: I think this topic (`validSettings`) can be removed.

sal_interfaces/Scheduler/Scheduler_Events.xml
andrewbheyer: I believe this was changed to 1 because a sting size of 1 dynamically grows to the size it needs to be.

sal_interfaces/Scheduler/Scheduler_Events.xml
andrewbheyer: Ok I will remove this topic.

sal_interfaces/Scheduler/Scheduler_Events.xml
tribeiro: Can you double check, just to make sure?

andrewbheyer: I have confirmation from Dave that the following statement is correct. 

Is it true then in the XML when you specify an <IDL_TYPE> to be "string" with an <IDL_SIZE> of "1" then this means the EFD will handle any size, given it is under the 30k byte size limit?

PR #52, opened 2018-12-04
TSS-3375: Include logging in BaseCsc

r-owen:


r-owen:
Based on private comments I updated the <Version> tag for Test, Script and ScriptQueue to 3.8.0 which is at least approximately the minimum SAL version needed.

dmills-lsst:
Can we clarify the relationship between these log events,and the generic errorCode message.
Also every event has an associated priority which could be made coincident with the level here.
ie A specific logevent which might be useful for a particular CSC's to monitor, but could be accompanied by 
general purpose log message for operator display etc, but any priority levels should be consistent.
Will we duplicate all errorCode messages, or should we make log messages non-error only and have the "watcher" deal with both sets.

r-owen:
An excellent question. `errorCode` has approximately the fields we need for log message output (and having a separate field for traceback is a nice touch). However, I am not comfortable using the `errorCode` field to mean log level (message severity); I believe it is intended to be an integer that tells you which error has occurred, which is not something I propose to try to support with logging.

As a result, my inclination is to have both events, but add `traceback` as a new field to the `logMessage` event.

Another option is to add a `level` field to `errorCode` and (preferably) document that it uses the Python logging severity levels. If we also document that an `errorCode` of 0 means "unspecified" then I think we have a reasonable solution. But I worry that changing an existing topic will raise havoc for existing CSCs.

CContaxis:
Keep it simple.

Just add a new event and be done. The errorCode event has a defined use 
case that a number of CSCs (M1M3 and M2 as an example) have already 
implemented. The errorCode event was meant to be the means for a CSC to 
determine the reason another CSC has entered the fault state.

There are also use cases for a CSC (not necessarily needing but maybe 
wanting to) write a log message over SAL other than "something really 
bad or unexpected happened".

These are 2 different use cases so keep them separate.

Chris

r-owen:
@dmills-lsst tells me he is fine adding the new `logMessage` logevent.

PR #55, opened 2018-12-12
TSS-3326: Shorten the salIndices and pastSalIndices fields

r-owen:


r-owen:
Sorry about that. I have shortened the 500 length topics to 400 and changed stopScripts to match (since I see no point to allowing stopping more scripts than can be reported as on the queue). I rebased the changes so if you have a checkout please resynchronize, e.g. using `reset --hard origin/tickets/TSS-3326`

PR #66, opened 2019-02-06
DM-17769: Restore Explanation fields to the original urls in ATDome XML

r-owen:
and move descriptive text to comments


jbuffill:
Ref: [https://jira.lsstcorp.org/browse/DM-17768](url)

PR #67, opened 2019-02-08
DM-17609: Improve ATPneumatics XML

r-owen:


r-owen:
Good catch. I had copied the new events from the commands of the same name (the name works better for an event than a command, but I felt it best to leave the command alone) and brought along a lot of extra unwanted fields. I made the new events look like the other events.

PR #71, opened 2019-02-22
Renamed attribute 'localTime' to 'localTimeString'.

ayayoshimura:
Renamed attribute 'localTime' to 'localTimeString' to avoid using MySQL reserved word LOCALTIME.


jbuffill:
Ref: https://jira.lsstcorp.org/browse/TPC-153

PR #73, opened 2019-03-05
TPC-158 Update ATPtg/MTPtg sal interface for az/rot current wrap and stopTracking command

ayayoshimura:
New command stopTracking and events ptgAzCurrentWrap and ptgRotCurrentWrap added. Wraps enumeration definition is updated. These updates are for TPC-158.


sal_interfaces/ATPtg/ATPtg_Commands.xml
```diff
@@ -917,4 +917,16 @@ fixed   = maintain sky orientation ( = AZEL)
...
\ No newline at end of file
+</SALCommand>
+<SALCommand>
+    <Subsystem>ATPtg</Subsystem>
+    <Version>3.8.41</Version>
+    <Author>pbt</Author>
+    <EFDB_Topic>ATPtg_command_stopTracking</EFDB_Topic>
+    <Alias>stopTracking</Alias>
+    <Device></Device>
+    <Property></Property>
```
tribeiro: I think this topic needs to have some attribute definition inside it, or it won't be a valid topic. You should probably add something like:

```
    <item>
        <EFDB_Name>stopTracking</EFDB_Name>
        <Description></Description>
        <IDL_Type>boolean</IDL_Type>
        <Units></Units>
        <Count>1</Count>
    </item>
```

Same goes to `MTPtg_command_stopTracking`

ayayoshimura: I found more commands that had no attribute. I've added 'ignored' boolean attribute to those commands. All commands now should have at least one attribute.

ayayoshimura:
I've made more changes since I've created the pull request. There is new DetailedStates azel. trackPosting event is added. raDecTarget command's ra and declination are changed to double attributes.

PR #101, opened 2019-07-01
DM-20327 Build/Deployment framework design

tribeiro:
Add Jenkinsfile and Dockerfile to run unit tests in CI.


tribeiro:
> On Line 24 of the Jenkinsfile, the version is hardcoded, in this case 3.10.2. It would be better if this value could be read from the VERSION file in the repo. This will reduce the number of changes needed to accomplish a new release.

Great idea.... I finally managed to make it work!

PR #121, opened 2019-07-29
Feature/andrewheyer/dm 20803

andrewbheyer:
This is the feature to have the Simulator column added to the XML. 

Here is the link to how the table looks https://ts-xml.lsst.io/v/feature-andrewheyer-DM-20803/index.html


awclemen:
Looks good - can we require & add "yes" or "no" to the column?  Also can we specify which ones we know have simulators (like MTM1M3, MTM2, etc) since we are at it?  Tiago could probably help you out with this.  Thanks.

PR #123, opened 2019-08-02
Feature/rbovill/report

rbovill:
These are improvements to the Jenkinsfile that allow the test results to be published properly.  The ${WORKSPACE}/robot directory needs to be cleared and set with proper permissions and then mounted into the Docker Container.  Finally, the link the report file itself was corrected.  This allows it to show on the Jenkins build page.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,9 @@ pipeline {
         stage("Clone robotframework_ts_xml") {
             steps {
                 sh """
-                rm -rf ${WORKSPACE}/rbtxml
+                rm -rf ${WORKSPACE}/rbtxml ${WORKSPACE}/robot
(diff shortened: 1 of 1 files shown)


tribeiro:
I think it's not posting the results when it fails because the step that copies the files are skipped.

```
Pipeline] { (Copy tests results)
Stage "Copy tests results" skipped due to earlier failure(s)
```

I think the stage 

```
        stage("Copy tests results") {
            steps {
                script {
                    sh """
                    docker cp ${container_name}:/home/appuser/Reports ${WORKSPACE}/robot
                    """
                }
            }
        }
```

Should be moved to post-processing steps.

rbovill:
These changes, plus additional work, were done in a different PR.

PR #136, opened 2019-09-17
DM-21247: add checkpoints to the ScriptQueue add and Script configure commands

r-owen:


r-owen:
I also took the liberty of cleaning up some things in SALGenerics, ATDome, ATDomeTrajectory, ATMCS and Test -- especially adding missing Description fields

PR #140, opened 2019-09-20
DM-20740: Implement FiberSpectrograph CSC version 1

parejkoj:
Update FiberSpectrograph XML to reflect changes needed for initial CSC implementation.

### DIFF
--- a/sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
+++ b/sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
@@ -1,39 +1,58 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsl"?>
-<SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
-<SALCommand>
-    <Subsystem>FiberSpectrograph</Subsystem>
-    <Version>3.7.1</Version>
-    <Author></Author>
-    <EFDB_Topic>FiberSpectrograph_command_captureSpectImage</EFDB_Topic>
-    <Alias>captureSpectImage</Alias>
-    <Device></Device>
-    <Property></Property>
-    <Action></Action>
-    <Value></Value>
-    <Description>Capture a spectrum for integration time specified</Description>
-    <item>
-        <EFDB_Name>integrationTime</EFDB_Name>
-        <Description>The detector in the spectrometer can capture light for a length of time specified by integration time</Description>
-        <IDL_Type>float</IDL_Type>
-        <Units>second</Units>
-        <Count>1</Count>
-    </item>
-    <item>
-        <EFDB_Name>imageType</EFDB_Name>
-        <Description>Type of image. For example, Standard exposure.</Description>
-        <IDL_Type>string</IDL_Type>
-        <IDL_Size>256</IDL_Size>
-        <Units>unitless</Units>
-        <Count>1</Count>
-    </item>
-    <item>
-        <EFDB_Name>lamp</EFDB_Name>
-        <Description>Light used for measurement.</Description>
-        <IDL_Type>string</IDL_Type>
-        <IDL_Size>256</IDL_Size>
-        <Units>unitless</Units>
-        <Count>1</Count>
-    </item>
-</SALCommand></SALCommandSet>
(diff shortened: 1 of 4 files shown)


parejkoj:
@rbovill Thanks for finding that. In the future, can you please comment on the exact line of code? It took me a while to find what you were referring to.

sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
```diff
@@ -1,39 +1,51 @@
...
+        <Subsystem>FiberSpectrograph</Subsystem>
+        <Version>3.7.1</Version>
+        <Author/>
+        <EFDB_Topic>FiberSpectrograph_command_cancelExposure</EFDB_Topic>
+        <Alias>cancelExposure</Alias>
+        <Device/>
+        <Property/>
+        <Action/>
+        <Value/>
+        <Description>Cancel an active exposure from the spectrograph.</Description>
```
r-owen: What happens if there is no active exposure? Does the command succeed or fail? It'd be nice to say here. (I suggest "succeed" so as to avoid race conditions).

sal_interfaces/FiberSpectrograph/FiberSpectrograph_Telemetry.xml
```diff
@@ -1,39 +1,25 @@
...
+<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
+    <SALTelemetry>
+        <Subsystem>FiberSpectrograph</Subsystem>
+        <Version>3.7.1</Version>
+        <Author/>
+        <EFDB_Topic>FiberSpectrograph_temperature</EFDB_Topic>
+        <Description>Temperature of the FiberSpectrograph</Description>
+        <item>
+            <EFDB_Name>temperature</EFDB_Name>
+            <Description>Measured internal spectrograph temperature in degrees Celcius.</Description>
```
r-owen: I suggest dropping "in degrees Celsius" because it duplicates information in the units -- here and elsewhere

sal_interfaces/SALSubsystems.xml
```diff
@@ -1033,20 +1033,20 @@
 
 <Subsystem>
   <Name>FiberSpectrograph</Name>
-  <Description></Description>
-  <Enumeration></Enumeration>
+  <Description>A CSC that manages the Fiber Spectrographs that are used to calibrate the tuneable lasers.</Description>
```
r-owen: @patrickingraham Is the fiber spectrograph used for on AuxTel used for a tunable laser? I thought we only used a tunable laser for the main telescope.

sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
```diff
@@ -1,39 +1,51 @@
...
+            <EFDB_Name>type</EFDB_Name>
+            <Description>Type of image. What does this mean?????</Description>
+            <IDL_Type>string</IDL_Type>
+            <IDL_Size>256</IDL_Size>
+            <Units>unitless</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>lamp</EFDB_Name>
+            <Description>Light used for measurement. Is this relevant???</Description>
```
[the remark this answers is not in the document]
patrickingraham: You're correct that this is essentially an "IMG_TYPE" keyword. A config file isn't the correct place for it as it can change from exposure-to-exposure.
We have to do a little more testing of the hardware to figure out where these will be mounted. I think it's best to leave it as a user-input for now (like IMG_TYPE with the take images command). In fact, we ought to think of just calling it IMG_TYPE.
John and myself need to have a talk about meta-data, this is the type of thing that will be discussed.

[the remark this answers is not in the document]
r-owen: @patrickingraham Isn't the field named `type` the field that you are talking about (the one that maps to IMGTYPE)

I was suggesting we keep the field named `type` and get rid of the field named `lamp` -- making *that* data based on which spectrograph is in use.

sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
parejkoj: I've written the spectrograph python library such that `cancel` when there is no exposure is a no-op. I'll note that here.

parejkoj:
I have updated the `expose` Command parameters based on our conversation with @patrickingraham . I'm waiting to hear a few things from Tim now.

sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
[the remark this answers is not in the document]
parejkoj: We clarified this in an offline conversation.

parejkoj:
@rbovill : I think you are still blocking the merge.

PR #145, opened 2019-09-27
DM-21437: Match up spectrograph index to serial number

parejkoj:
Update the device names, and remove a bunch of files that should never have been committed.


.gitignore
```diff
@@ -1 +1,14 @@
...
+.sconsign.dblite
+config.log
+.sconf_temp
+*.o
+*.os
+*.so
+*.cfgc
+*.pyc
+*_wrap.cc
+*Lib.py
```
r-owen: I don't think we should expect to find any of the above files in ts_xml except possibly lines 2-4 (.scon... and config.log) and not even those if we can ditch the SConstruct file.

PR #146, opened 2019-09-30
Feature/cap 209

tony-johnson:
ATCamera changes for writing FITS headers (CAP-209) and resync MTCamera commands/events.


rbovill:
I addressed the issue that was the Jenkins build to fail, at least temporarily.  I was working on an improvement to the Jenkinsfile and it somehow made it into the develop branch prematurely.  When @tonyj rebased, he picked up that change and that caused the issue.  I removed the errant change from the Jenkinsfile on the the feature/CAP-209 branch.

PR #153, opened 2019-10-18
Davemills/dm 18424

dmills-lsst:
https://jira.lsstcorp.org/browse/DM-18424


dmills-lsst:
I have removed the settingsApplied events from each CSC. In cases where they had useful
content they have been renamed instead. The CSC developers still need to implement the 
code to announce their specific settingsAppliedBlah events

PR #162, opened 2019-11-05
Add stop command to MTMount

ander-ansuategi:


andrewbheyer:
It appears that the build is failing due to telemetry not complying to the current naming convention standards seen here https://ts-xml.lsst.io/v/tickets-dm-20234/sal_constraints_and_recommendations.html. I believe Tekniker is grandfathered in and won't need to update the XML. @rbovill can you confirm if this is true or not?

rbovill:
The MTMount XML, as it currently is released, does not need to be updated to conform to the new standards.  Since we are making changes to the Commands XML file, we should also update fields to conform to the standards.  If there is a problem with that, I can update the tests to work around the failure.

rbovill:
The problems with this update include blank Descriptions, blank Units, and Degrees and Seconds are not valid units.  For the blank Descriptions, some text needs to be there.  Similar for the blank units, but valid units are necessary.  

For explanations of valid units, see [SAL standards](https://ts-xml.lsst.io/v/tickets-dm-20234/sal_constraints_and_recommendations.html#sal-topic-and-parameter-description-and-unit-fields) and the linked [Astropy units definitions](http://docs.astropy.org/en/stable/units/index.html#module-astropy.units.si).

PR #176, opened 2019-11-19
CAP-356: Rename ATDomeTrajectory `settingsApplied` event to `algorithm`

r-owen:


tribeiro:
I would suggest renaming it to `settingsAppliedAlgorithm`...

mareuter:
I'm going to reverse my decision and agree with Tiago. In the absence of information and looking at topic names a list, there would be no way to know that this is a settings applied type event by looking at the name. One would have to consult the EFD for the list from the _settingsApplied_ event.

r-owen:
I don't understand why this matters. `algorithm` is an event. Like all events it is output when it changes. In this case that what changes it is the `start` command, but that's an internal detail.

tribeiro:
can we merge this?

mareuter:
As long as the event is listed in the settingsApplied event, I'm fine with the name. It can be merged.

r-owen:
My plan is to add support for the generic settingsApplied event to ts_salobj. Until then ts_ATDomeTrajectory will not output it (well, until I merge this it outputs a non-compliant version).

There's a bit of chicken-and-egg going on here. I see two paths for getting a generic event, neither of which is going to happen very soon (if only because we need to be sure we are switching to my Python CSCs for hexapod and rotator before we can start messing with that XML):
* Piecemeal eliminate the custom settingsApplied event for *all* SAL components. Once that is done it is safe to add it as a generic event.
* Release ts_sal 4.1 and then modify SALSubsystems.xml to explicitly list the generic events that each CSC uses. At that point it is safe to add any generic events we like. I am looking forward to doing this regardless of whether we hold up settingsApplied on it. I admit that an *accurate* list of generic events for each CSC will be somewhat difficult, but it's easy enough to just list all the current generic events and then prune as folks have time.

I'm really not keen on settingsAppliedXXX events. I'll rename this if I'm forced to, but I'd much rather leave this code as it is.

mareuter:
I'm attacking the problem via the first bullet. As I said before, I'm fine with your name. The settingsApplied{name} is a recommendation, not a requirement. Please go ahead and finish this off.

PR #177, opened 2019-11-20
Add runtime language tags to SALSubsystems.xml

dmills-lsst:
Also added SALSubsystems schema


rbovill:
Additionally, does this schema file override the old one?  If so, that schema file should be removed.

rbovill:
In SALSubsystemSet.xsd, by defining the SubsystemType as an <xsd:choice>, we are limiting the entries to only ONE OF the defined elements.  Since we want all of them, as options at least, then we need to define the SubsystemType as an <xsd:sequence>.

Additionally, with that change, there are still several errors, as <Version> and <Simulator> are set as minOccurs=1, so the blank entries are failing.  We should discuss how best to handle this.  Change minOccurs to 0?  Write tickets for failing subsystems?

PR #180, opened 2019-11-25
Ticket/dm22032

rbovill:
Initial effort to convert the Robot-Framework tests into pytest unit tests.


tests/Unit_Validator.py
```diff
@@ -0,0 +1,16 @@
...
+	def check_unit(self, unit_str):
+		try:
+			return astropy.units.Quantity("1 " + unit_str)
+		except ValueError:
+			return "ValueError: " + unit_str + " is not a valid unit."
+		except:
+			return "UnknownError: " + unit_str
+
+if __name__ == "__main__":
+	Unit_Validator.main()
```
r-owen: How does this work? I would not expect `Unit_Validator` to have a `main` because it is a subclass of `object`. Did you mean to make it a subclass of `unittest.TestCase`?

tests/Unit_Validator.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import astropy.units
+
+class Unit_Validator():
```
r-owen: Please rename to `Unit_Validator` (I think pep8 checkers will complain otherwise) and remove the `()` if it's really not a subclass.

tests/Unit_Validator.py
```diff
@@ -0,0 +1,16 @@
...
+
+	def check_unit(self, unit_str):
+		try:
+			return astropy.units.Quantity("1 " + unit_str)
+		except ValueError:
+			return "ValueError: " + unit_str + " is not a valid unit."
+		except:
+			return "UnknownError: " + unit_str
```
r-owen: This test will do nothing because it has no method whose name starts with `test`. Is there some code that should call check_unit?

tests/test_Alias.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import pytest
+import xml.etree.ElementTree as ET
+import xml_common
```
r-owen: Please do not import anything from the test directory.

Move common code to `python/lsst/ts/xml/`. I suggest a module name of `test_utils.py` because we use that a lot for this kind of thing. Naming the module `xml_common` would not be appropriate because all ts_xml library code is, by definition, common to all of the ts_xml package.

This allows the tests to be run from any directory and also makes the code easier to follow, in that common code is in the standard location for library modules.

tests/Unit_Validator.py
rbovill: Unit_Validator isn't a test, it's a utility called by test_Units.py.  I guess the check_unit function should go into test_utils.py?

tests/Unit_Validator.py
rbovill: Completely redid this implementation.  The check_unit() function is now in test_utils.py.

tests/Unit_Validator.py
rbovill: Completely redid this implementation. The check_unit() function is now in test_utils.py.

tests/test_Alias.py
rbovill: Moved xml_common.py in tests/ to python/lsst/ts/xml/test_utils.py.

tests/test_Generics.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_salgenerics_commands():
+	"""Test that SALGenerics.xml defines the expected set of generic commands.
+
+	Parameters
+	----------
+	sal_generics_file: `pathlib.Path(sal_interfaces/SALGenerics.xml)`
+		The file being tested
+	generic_commands : `test_utils.generic_commands`
+		The list of Generic Commands
```
r-owen: Please delete the parameters section (because this function takes no parameters).

tests/test_Generics.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_salgenerics_events():
+	"""Test that SALGenerics.xml defines the expected set of generic events.
+
+	Parameters
+	----------
+	sal_generics_file: `pathlib.Path(sal_interfaces/SALGenerics.xml)`
+		The file being tested
+	generic_events : `test_utils.generic_events`
+		The list of Generic Events
```
r-owen: Again...no parameters.

tests/test_Generics.py
```diff
@@ -0,0 +1,114 @@
...
+	"""Test that CSC XML files do not define any of the generic topics.
+
+	NOTE: Telemetry is skipped because there is no generic telemetry.
+
+	Parameters
+	----------
+	generic_commands : `test_utils.generic_commands`
+		The list of Generic Commands
+	generic_events : `test_utils.generic_events`
+		The list of Generic Events
```
r-owen: This function doesn't take the above 2 parameters.

tests/test_NoReservedWords.py
```diff
@@ -0,0 +1,38 @@
...
+	jira = check_for_issues(csc, topic)
+	if jira:
+		pytest.skip(jira + ": " + str(xmlfile.name) + " <EFDB_Name> is not properly formed.")
+	# Test the <EFDB_Name> fields do not use Reserved Words.
+	with open(str(xmlfile), "r", encoding="utf-8") as f:
+		tree = ET.parse(f)
+	root = tree.getroot()
+	for name in root.findall(f"./{saltype}/item/EFDB_Name"):
+		assert name.text.upper() not in set(ts_xml.idl_reserved + ts_xml.db_reserved), \
+			'Reserved Word ' + name.text + ' used one or more times.'
```
r-owen: Please consider accumulating a list of violations, so the user see and fix all errors at once.

(What I had in mind is to use the loop to accumulate bad names in a list. Then after the loop assert that the list is empty.)

tests/test_SALSubsystems.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.mark.parametrize("root,csc,simulator", get_csc_simulator())
+def test_simulator_tag(root, csc, simulator):
+	"""Test that the <Simulator> tag is correctly defined for each CSC.
+
+	Parameters
+	----------
+	sal_subsystems_file: `pathlib.Path(sal_interfaces/SALSubsystems.xml)`
+		Full filepath to SALSubsystems.xml dictionary file.
+	csc : `test_utils.subsystems`
+		Name of the CSC
```
r-owen: This list doesn't match the actual arguments (though csc does).

tests/test_TopicNaming.py
```diff
@@ -0,0 +1,136 @@
...
+		assert name.text.split('_')[index] == topictype, \
+			"<EFDB_Topic> " + name.text + ' in ' + str(xmlfile.name) + \
+			' does not properly contain the topicType string.'
+
+
+@pytest.mark.parametrize("xmlfile,csc,topic", ts_xml.get_xmlfile_csc_topic())
+def test_topic_naming_alias(xmlfile, csc, topic):
+	"""Test that the <EFDB_Topic> field for topics is properly formed.
+	The <EFDB_Topic> is a compound word comprising the CSC Name, the topic type
+	and the topic name.  The name begins with a lowercase letter and contains
```
r-owen: This is not quite right for telemetry. Perhaps something like "the topic type, if not Telemetry, and the topic name"?

tests/test_TopicNaming.py
```diff
@@ -0,0 +1,136 @@
...
+	only alphanumeric and underscore characters.
+
+	This tests the third root, the topic name.
```
r-owen: Not quite right for telemetry. One possibility is "the last" instead of "the third".

r-owen: Consider also testing that the first element matches the CSC name and the second element (if not Telemetry) matches the topic. (Or perhaps you already do that in another test and I missed it).

tests/test_Generics.py
rbovill: Fixed

tests/test_Generics.py
rbovill: Fixed

tests/test_SALSubsystems.py
rbovill: Fixed

tests/test_TopicNaming.py
rbovill: Changed "third" to "last."  The other tests are in fact done in the previous test cases of this module.

tests/test_TopicNaming.py
rbovill: Updated the descriptions to

> Test that the <EFDB_Topic> field for topics is properly formed.
    The <EFDB_Topic> is a compound word comprising the CSC Name, 
    the topic type, if applicable, and the topic name.
    Names begin with a lowercase letter and contain only alphanumeric
    and underscore characters.

The last sentence, "Names being with..." is omitted in the test for the topic type.

tests/test_Generics.py
rbovill: Done

setup.cfg
```diff
@@ -0,0 +1,20 @@
+[aliases]
+test=pytest
+
+[flake8]
+max-line-length = 110
+max-doc-length = 79
+ignore = E117, E133, E226, E228, N802, N803, N806, N812, N813, N815, N816, W191, W391, W504
```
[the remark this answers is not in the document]
rbovill: Done

setup.cfg
```diff
@@ -0,0 +1,20 @@
...
+exclude =
+  doc/*,
+  python/lsst/ts/xml/test_utils.py,
```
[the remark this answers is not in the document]
rbovill: Fixed.

setup.cfg
```diff
@@ -0,0 +1,20 @@
...
+  sal_interfaces/*,
+  schema/*,
+  scripts/*,
+  tests/.tests,
+  __init__.py,
+  setup.py
+
+[tool:pytest]
+addopts = --flake8 --ignore=scripts --ignore=doc --ignore=setup.py --ignore=python/lsst/ts/xml/test_utils.py --ignore=python/lsst/ts/xml/__init__.py
```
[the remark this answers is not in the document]
rbovill: I fixed --ignore clauses and testutils.py.  However, I am not getting the N814 camelcase warning on my end.  Different version of pytest or flake8?

.gitignore
```diff
@@ -9,6 +9,9 @@ config.log
 *.pyc
 *_wrap.cc
 *Lib.py
+python/ts_xml.egg-info/*
+.eggs/*
+.pytest_cache/*
```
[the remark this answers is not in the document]
rbovill: Done

tests/test_NoReservedWords.py
rbovill: Done.

setup.cfg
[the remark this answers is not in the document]
r-owen: Probably. I'm using the one provided with one of Tiago's docker images, so it'll be widely used. I strongly encourage you to make this change.

I can run things again to make sure they all pass for me (or Tiago can; presumably he'll see the same thing) if you like.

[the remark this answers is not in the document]
rbovill: I changed the xml.etree import to import as et instead of ET.  This fixes the issue.

PR #189, opened 2019-12-11
Feature/andrewheyer/dm 22382

andrewbheyer:


rbovill:
Unit tests for the new <Configurable> tag should be made before merging this PR.

andrewbheyer:
I made a slight deviation to how the configurable test is done. When the following definition is called 
`def get_csc_simulator():
    root = get_file_root_element()
    arguments = []
    for csc in ts_xml.subsystems:
        simulator = root.find("./SALSubsystem/[Name='" + csc + "']/Simulator").text
        arguments.append((root, csc, simulator))
    return arguments`
I would receive the following feedback for XML's that did not have a <Configurable> tag 
`E   AttributeError: 'NoneType' object has no attribute 'text'`

I may very well be running the test incorrectly, to produce more detailed exceptions I first assert in my test the tag can be found, then assert that the tag is either "yes" or "no". This produces the following when the tags are not found
`E           AssertionError: ATAOS <Simulator> tag is NOT defined.`
and the following if not "yes" or "no"
`E           AssertionError: ATAOS <Simulator> text must either be 'yes' or 'no'`

rbovill:
The configurable tag test needs to use the parametrize decorator.  also, the first check should model after the test_simulator_tag test above it, to verify the tag is a proper element instead of not None, as that test will fail if the tag is not defined.

PR #195, opened 2019-12-16
DM-22627 replace duration with time in Script and Test XML

r-owen:


r-owen:
Cancelling this pull request for now because we may decide we don't need the rename.

PR #196, opened 2019-12-16
DM-22628: Rename duration to timespan for Watcher mute command.

r-owen:


r-owen:
We have decided to delay this change until we decide if it is necessary.

PR #197, opened 2019-12-17
Tickets/dm22231

rbovill:
*  Updated the db_reserved list in testutils.py to the list of InfluxDB words. Added Jira skip tags to now failing tests in test_NoReservedWords.py.


tests/test_NoReservedWords.py
```diff
@@ -6,7 +6,42 @@
...
+        jira = "DM-22623"
+    elif csc == "PromptProcessing" and (topic == "Telemetry" or topic == "Events"):
+        jira = "DM-22624"
+    elif csc == "Scheduler" and topic == "Telemetry":
+        jira = "DM-22625"
+    elif csc == "Script" and topic == "Events":
+        jira = "DM-22626"
+    elif csc == "Test" and topic == "Commands":
+        jira = "DM-22627"
+    elif csc == "Watcher" and (topic == "Commands" or topic == "Events"):
```
r-owen: The only Watcher error I know of is in Commands (the mute command).

tests/test_NoReservedWords.py
```diff
@@ -6,7 +6,42 @@
 
 
 def check_for_issues(csc, topic):
-    jira = ""
+    if csc == "ATAOS" and topic == "Commands":
+        jira = "DM-22612"
+    elif csc == "ATMCS" and (topic == "Commands" or topic == "Events"):
```
r-owen: I am not suggesting you change the code, but in future you might consider reducing repetition using:
```
...and topic in ("Commands", "Events"):
```

tests/test_NoReservedWords.py
rbovill: Watcher Events uses NAME, which is also an InfluxDB Reserved Word

tests/test_NoReservedWords.py
rbovill: I will keep that in mind for next time.

PR #198, opened 2019-12-17
tickets/cap396

rbovill:
Added a workaround for the Torr and mTorr units to test_Units.py.

### DIFF
--- a/tests/test_Units.py
+++ b/tests/test_Units.py
@@ -50,7 +50,9 @@ def test_units(xmlfile, csc, topic):
     for unit in root.findall(f"./{saltype}/item/Units"):
         if not unit.text:
             assert False, 'Units cannot be blank.'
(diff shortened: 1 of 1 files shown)


tests/test_Units.py
```diff
@@ -50,7 +50,9 @@ def test_units(xmlfile, csc, topic):
     for unit in root.findall(f"./{saltype}/item/Units"):
         if not unit.text:
             assert False, 'Units cannot be blank.'
-        elif unit.text == "unitless" or unit.text == "dimensionless":
+        elif unit.text in ("unitless", "dimensionless"):
+            assert True
+        elif unit.text in ("Torr", "mTorr"):
```
timj: Can we add a comment along the lines of
```python
elif unit.text in (blah):  # Check to be removed when DM-nnnn is done
```
?

colinwinslow: Sure.

PR #200, opened 2019-12-18
Split the Reserved Word tests into three parts

rbovill:
The Reserved Word tests now check IDL, DB Critical and DB Optional as three separate tests.  As DB optional words become critical, simply remove the words from the optional list and add them to the critical list in testutils.py


rbovill:
Implemented the control function.  What do you think?

PR #201, opened 2019-12-19
DM-21699: Clean up the Rotator and Hexapod XML

r-owen:


teweitsai:
In the previous integration meeting, Doug had pointed out that he did want to be involved in the discussion of xml change for rotator and hexapod. In addition, since the product owner is @bxin, it might be good to have the discussion with him together for the change of xml interface.

I also have some questions for the xml changes as well. I am not sure what should be the update of wrapper code to support some xml change you proposed. Thanks!

teweitsai:
I added @andrewbheyer in reviewer as well because the rotator xml change might affect the CCW (part of MTMount) as well. For example, the updated rotator xml telemetry will affect the CCW LabVIEW code.

r-owen:
It would be helpful to know which events the LabVIEW code is relying on. It should not be listening to Rotator commands because the Rotator might reject those (e.g. if it halts for some reason it will stop accepting track commands, but somebody may still try to send them).

r-owen:
Here is a description of the changes, with an explanation for each. I suggest you comment there: https://confluence.lsstcorp.org/display/LTS/Rotator+XML+Changes

sal_interfaces/Rotator/Rotator_Commands.xml
```diff
@@ -3,46 +3,6 @@
 <SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
 
-<SALCommand>
-    <Subsystem>Rotator</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Rotator_command_configureAcceleration</EFDB_Topic>
```
bxin: See comments on the Hexapod configure commands. I think this should stay.

sal_interfaces/Rotator/Rotator_Commands.xml
```diff
@@ -3,46 +3,6 @@
...
-        <Units>deg/s2</Units>
-        <Count>1</Count>
-    </item>
-</SALCommand>
-
-<SALCommand>
-    <Subsystem>Rotator</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Rotator_command_configureVelocity</EFDB_Topic>
```
bxin: same as above.

sal_interfaces/Rotator/Rotator_Commands.xml
```diff
@@ -164,73 +97,6 @@
...
-        <Units>deg</Units>
-        <Count>1</Count>
-    </item>
-</SALCommand>
-
-<SALCommand>
-    <Subsystem>Rotator</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Rotator_command_moveConstantVelocity</EFDB_Topic>
```
bxin: According to the engineering team, this command is useful for testing. So we should keep this.

sal_interfaces/Rotator/Rotator_Commands.xml
```diff
@@ -164,73 +97,6 @@
...
-        <Units>dimensionless</Units>
-        <Count>1</Count>
-    </item>
-</SALCommand>
-
-<SALCommand>
-    <Subsystem>Rotator</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Rotator_command_velocitySet</EFDB_Topic>
```
bxin: same comment as above. I am OK with merging this and the above into one command.

sal_interfaces/Rotator/Rotator_Events.xml
```diff
@@ -110,78 +110,54 @@
...
+        <IDL_Type>double</IDL_Type>
+        <Units>second</Units>
         <Count>1</Count>
     </item>
 </SALEvent>
 <SALEvent>
     <Subsystem>Rotator</Subsystem>
     <Version>3.5.1</Version>
     <Author></Author>
-    <EFDB_Topic>Rotator_logevent_deviceError</EFDB_Topic>
```
bxin: Same as the deviceError event for hexapod. I prefer to keep this if it is not too much trouble.

sal_interfaces/Rotator/Rotator_Events.xml
```diff
@@ -265,37 +241,6 @@
         <Count>1</Count>
     </item>
 </SALEvent>
-<SALEvent>
-    <Subsystem>Rotator</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Rotator_logevent_rejectedCommand</EFDB_Topic>
```
bxin: Same comment as for Hexapod rejectedCommand.

sal_interfaces/Hexapod/Hexapod_Events.xml
```diff
@@ -122,62 +122,39 @@
...
+        <IDL_Type>boolean</IDL_Type>
+        <Units>unitless</Units>
+        <Count>6</Count>
     </item>
 </SALEvent>
 <SALEvent>
     <Subsystem>Hexapod</Subsystem>
     <Version>3.5.1</Version>
     <Author></Author>
-    <EFDB_Topic>Hexapod_logevent_deviceError</EFDB_Topic>
```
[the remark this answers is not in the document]
r-owen: The problem with deviceError, as implemented by the vendor, is that you get one event per flag bit when multiple bits are set, and no way to know when the bits are no longer set. The replacement event solves both of these problems by outputting the data as a bitmask. There are constants in the XML file that tell you what each bit means.

sal_interfaces/Hexapod/Hexapod_Events.xml
```diff
@@ -480,37 +457,6 @@
         <Count>1</Count>
     </item>
 </SALEvent>
-<SALEvent>
-    <Subsystem>Hexapod</Subsystem>
-    <Version>3.5.1</Version>
-    <Author></Author>
-    <EFDB_Topic>Hexapod_logevent_rejectedCommand</EFDB_Topic>
```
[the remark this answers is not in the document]
r-owen: The new CSC uses the standard mechanism: the command is acknowledged as failed.

sal_interfaces/Rotator/Rotator_Commands.xml
r-owen: It is not supported by the CSC. Instead I offer a command-line script that allows you to command a ramp, which has the same effect.

sal_interfaces/Hexapod/Hexapod_Commands.xml
```diff
@@ -7,18 +7,18 @@
     <Subsystem>Hexapod</Subsystem>
     <Version>3.5.1</Version>
     <Author></Author>
-    <EFDB_Topic>Hexapod_command_configureAcceleration</EFDB_Topic>
```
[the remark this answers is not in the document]
r-owen: OK

sal_interfaces/Rotator/Rotator_Events.xml
r-owen: I'm afraid it's not going to happen, for the reasons given above.

sal_interfaces/Rotator/Rotator_Commands.xml
r-owen: See above.

sal_interfaces/Rotator/Rotator_Commands.xml
bxin: @tribeiro pointed me to this https://github.com/lsst-ts/ts_rotator/blob/00e19f5af74d83595bf104a908d6480b6b3095f2/python/lsst/ts/rotator/rotator_commander.py#L44-L47

r-owen: Exactly. Also the next command listed gives you a sinusoidal motion and we can add other shapes if they prove useful for testing.

sal_interfaces/Hexapod/Hexapod_Events.xml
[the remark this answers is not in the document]
r-owen: I forgot to mention but the replacement source of information is the `applicationStatus` field of the new `controllerStatus` event.

PR #204, opened 2019-12-19
Add "statusCode" and "description" to each ATArchiver and CCArchiver

srp3rd:
…event


mareuter:
Does MTArchiver require the same info too? If so, might as well add them.

srp3rd:
Added the XML changes for the MTArchiver

PR #215, opened 2020-01-08
DM-22819: Add logLevel event to LOVE

r-owen:


tribeiro:
Shouldn't this be a generic event?

mareuter:
I think the answer is, yes, it will be, but until we can whitelist supported generic events.

r-owen:
Exactly. In fact the current ts_sal does support whitelisting generic events, but the error handling is not where it should be. The next release of ts_sal will fix the error handling, at which point I want to update the XML for Script and LOVE to use generic topics.

PR #217, opened 2020-01-22
Story DM-23058: Created code that generates SAL XML for the HVAC syst…

wvreeven:
Created code that generates SAL XML for the HVAC system. This file is the first result of that code and certainly not the last. I verified that it doesn't break the unit tests though I am well aware that it probably doesn't entirely follow the naming conventions yet.


sal_interfaces/HVAC/HVAC_Telemetry.xml
```diff
@@ -1,256 +1,1010 @@
...
-<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
-<SALTelemetry>
+<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
+  <SALTelemetry>
     <Subsystem>HVAC</Subsystem>
-    <Version>3.8</Version>
-    <Author></Author>
-    <EFDB_Topic>HVAC_lowerAHUStatus</EFDB_Topic>
+    <EFDB_Topic>HVAC_topic_(mqtt)_(todas_variables_)</EFDB_Topic>
```
r-owen: The parenthesis are not allowed.

Note: I encourage you to run the SAL IDL generator on these files, as well as run the unit tests. In the development Docker image run `make_idl_file.py HVAC`. It is even more important for that to run correctly (and generate usable IDL files) that it is for the ts_xml unit tests to run, since it is the files generated by SAL that we actually use.

sal_interfaces/HVAC/HVAC_Telemetry.xml
```diff
@@ -1,256 +1,1010 @@
...
+  </SALTelemetry>
+  <SALTelemetry>
+    <Subsystem>HVAC</Subsystem>
+    <EFDB_Topic>HVAC_lsst_fwdsl_piso01_fwdsl_tcc-p1-sala-maquinas</EFDB_Topic>
+    <item>
+      <EFDB_Name>lsst_fwdsl_piso01_fwdsl_vin_01_fwdsl_estado_funcionamiento</EFDB_Name>
+      <Description>lsst_fwdsl_piso01_fwdsl_vin_01_fwdsl_estado_funcionamiento</Description>
+      <IDL_Type/>
```
r-owen: IDL_Type must be specified. This is the data type that DDS will use to transmit the data, so it is fundamental. (Perhaps DDS or the SAL IDL generator has a default, but if so, I have no idea what it might be).

I hope MQTT makes the information directly available (e.g. via some kind of introspection). But if not, then you can work it out based on the types of the fields in the MQTT messages. The problem with doing that is if MQTT reports the fields as standard Python data types (as opposed, say, to ctypes or numpy types) then that loses information: you can't tell the difference between different types of int, nor the difference between float and double.

The available XML data types are described here:
https://confluence.lsstcorp.org/display/LTS/SALPY+Interfaces
note that many of them map to the same data types in IDL, so there are fewer actual choices than it may look like at first.

sal_interfaces/HVAC/HVAC_Telemetry.xml
```diff
@@ -1,256 +1,1010 @@
...
+    <item>
+      <EFDB_Name>alarma_filtro</EFDB_Name>
+      <Description>alarma_filtro</Description>
+      <IDL_Type/>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>reset_alarma</EFDB_Name>
+      <Description>reset_alarma</Description>
```
r-owen: Any idea what this is? It sounds more like a command than telemetry.

sal_interfaces/HVAC/HVAC_Telemetry.xml
wvreeven: That one slipped my attention. I improved my code because this one is not supposed to be in the XML. Also I improved the TopicNaming test to catch this kind of naming errors.

sal_interfaces/HVAC/HVAC_Telemetry.xml
wvreeven: I made sure that (for now) all IDL_Types are set to int, until I receive the proper type info from DatControl.

sal_interfaces/HVAC/HVAC_Telemetry.xml
wvreeven: To be honest, I don't know but according to the information that was given to me this is a genuine Telemetry topic.

wvreeven:
I pushed new changes. Can you please review again?

tests/test_TopicNaming.py
```diff
@@ -129,6 +129,6 @@ def test_topic_naming_alias(xmlfile, csc, topic):
             index = 1
         else:
             index = 2
-        assert re.match(r"^[a-z]([a-zA-Z0-9_]*$)", name.text.split('_')[index]) is not None, \
+        assert re.match(r"^[a-z]([a-zA-Z0-9_]*$)", name.text.split('_', index)[index]) is not None, \
```
r-owen: Please consider providing the argument name in split, to make it clearer what the index is doing as an argument, in other words:
```
... name.text.split('_', maxsplit=index)[index])`
```

wvreeven: Yes I added that too now. As for the proper field types: I can guess based on the telemetry received from the MQTT server but I prefer to do all of this in one go once we receive the information from DatControl since there will be big changes then anyway.

PR #221, opened 2020-01-24
Added a new IDL Type test and modified the testutils.py file to conta…

wvreeven:
…in a list of allowed IDL Types.

### DIFF
--- a/python/lsst/ts/xml/testutils.py
+++ b/python/lsst/ts/xml/testutils.py
@@ -59,6 +59,11 @@ idl_reserved = [
     'VALUETYPE', 'VOID', 'WCHAR', 'WSTRING'
 ]
 
+"""Define the list of IDL Types"""
+idl_types = ['boolean', 'byte', 'octet', 'short', 'int', 'long', 'long long',
+             'unsigned short', 'unsigned int', 'unsigned long',
+             'unsigned long long', 'float', 'double', 'string', 'char']
+
 db_critical_reserved = [
     "TIME"
 ]
--- a/tests/test_IDL_Type.py
+++ b/tests/test_IDL_Type.py
@@ -35,10 +35,6 @@ def test_idl_type(xmlfile, csc, topic):
         tree = et.parse(f)
     root = tree.getroot()
     for idl_type in root.findall(f"./{saltype}/item/IDL_Type"):
-        if not idl_type.text:
-            assert False, 'IDL_Type cannot be blank.'
-        if idl_type.text not in ts_xml.idl_types:
-            assert False, 'IDL_Type "' + idl_type.text + '" needs to be one of ' + \
-                str(ts_xml.idl_types)
-        else:
-            assert True
+        assert idl_type.text is not None, 'IDL_Type cannot be blank.'
+        assert idl_type.text in ts_xml.idl_types, 'IDL_Type "' + idl_type.text + \
+            '" needs to be one of ' + str(ts_xml.idl_types)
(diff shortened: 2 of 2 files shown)


rbovill:
Instead of `    if not idl_type.text:
            assert False, 'IDL_Type cannot be blank.'
        if idl_type.text not in ts_xml.idl_types:
            assert False, 'IDL_Type "' + idl_type.text + '" needs to be one of ' + \
                str(ts_xml.idl_types)
        else:
            assert True`

How about `assert idl_type.text is not None, 'IDL_Type cannot be blank.'
assert idl_type.text in ts_xml.idl_types, 'IDL_Type "' + idl_type.text + \
            '" needs to be one of ' + str(ts_xml.idl_types)`

tests/test_IDL_Type.py
```diff
@@ -0,0 +1,44 @@
...
+    jira = check_for_issues(csc, topic)
+    if jira:
+        pytest.skip(jira + ": " + str(xmlfile.name) +
+                    " <IDL_Type> fields do not conform to IDL standards.")
+    # Test the attribute <Description> fields.
+    with open(str(xmlfile), "r", encoding="utf-8") as f:
+        tree = et.parse(f)
+    root = tree.getroot()
+    for idl_type in root.findall(f"./{saltype}/item/IDL_Type"):
+        if not idl_type.text:
```
rbovill:
```suggestion
        if not idl_type.text:
```

tests/test_IDL_Type.py
```diff
@@ -0,0 +1,44 @@
...
+            assert False, 'IDL_Type cannot be blank.'
+        if idl_type.text not in ts_xml.idl_types:
+            assert False, 'IDL_Type "' + idl_type.text + '" needs to be one of ' + \
+                str(ts_xml.idl_types)
+        else:
+            assert True
```
rbovill: Instead of multiple if statements, how just two assert statements:

>     assert idl_type.text is not None, 'IDL_Type cannot be blank.'
>     assert idl_type.text in ts_xml.idl_types, 'IDL_Type "' + idl_type.text + \
>            '" needs to be one of ' + str(ts_xml.idl_types)

wvreeven: Good idea, thanks.

wvreeven: Updated code pushed upstream. Please review again.

rbovill: Looks good.

wvreeven: Actually, the line
`assert idl_type.text is not None, 'IDL_Type cannot be blank.'`
can be replaced with
`assert idl_type.text, 'IDL_Type cannot be blank.'`

What do you think?

wvreeven: I discussed this with Rob and we agreed not to make that final change.

python/lsst/ts/xml/testutils.py
```diff
@@ -59,6 +59,11 @@
     'VALUETYPE', 'VOID', 'WCHAR', 'WSTRING'
 ]
 
+"""Define the list of IDL Types"""
+idl_types = ['boolean', 'byte', 'octet', 'short', 'int', 'long', 'long long',
+             'unsigned short', 'unsigned int', 'unsigned long',
+             'unsigned long long', 'float', 'double', 'string', 'char']
+
```
r-owen: For future reference: the "in" test would be more efficient if `idl_types` was a set, e.g. `idl_types = set(['boolean', ...])`

Efficiency is no big deal here as I imagine the test runs fairly quickly anyway.

wvreeven: Thanks for educating me!

PR #222, opened 2020-01-24
Added an additional assertion in test_AttributeDescription.py to catc…

rbovill:
…h blank <Description> fields better.  Modified second assertion to catch <Description> fields that only contain whitespace.


rbovill:
Fixed the Flake8 failures.

PR #226, opened 2020-01-31
Davemills/dm 22141

dmills-lsst:
Moved rpm-dep from ts_sal


dmills-lsst:
it should be up to date now

PR #234, opened 2020-02-12
Add automatically generated XML for comcam vacuum, quadbox and rebpow…

tony-johnson:
Initial attempt at automatic XML generation for ComCam (CAP-450). Appears to pass at pytest tests, and compiles and riuns all tests in simple-sal. Please include in 4.7 release.


mareuter:
What does LDO stand for? It's never explained in any of the descriptions. I assume that LDO2 is just a second instance of a LDO on the same board.

mareuter:
I'm not sure if the attributes in the CCCamera_quadbox_PDU_XXXX topics need to have XXXX repeated in their name, since they are namespaced by the topic name. I assume that this would take time to fix.

tony-johnson:
> I'm not sure if the attributes in the CCCamera_quadbox_PDU_XXXX topics need to have XXXX repeated in their name, since they are namespaced by the topic name. I assume that this would take time to fix.

They don't, and Homer said he could remove them, but it would require changing the groovy definition in CCS, and then regenerating the XML. We do not have time if we want this by 9am tomorrow.

tony-johnson:
> What does LDO stand for? It's never explained in any of the descriptions. I assume that LDO2 is just a second instance of a LDO on the same board.

From Owen: Google reminds me that it’s “low-dropout regulator”.  I don’t remember why, but some channels have an extra ADC for measuring the voltage.

mareuter:
Would you have an estimate on how long a change would take? I'm willing to delay a little bit since this release is largely for ComCam and the AuxTel stuff can't make the summit next week.

tony-johnson:
> Would you have an estimate on how long a change would take? I'm willing to delay a little bit since this release is largely for ComCam and the AuxTel stuff can't make the summit next week.

I have asked on #comcam-computing

PR #235, opened 2020-02-12
Add additional AuxTel WREB telemetry item for HVSwitch (CAP-423)

tony-johnson:
Please include in 4.7 release. Needed for CAP-423.

### DIFF
--- a/sal_interfaces/ATCamera/ATCamera_Telemetry.xml
+++ b/sal_interfaces/ATCamera/ATCamera_Telemetry.xml
@@ -239,6 +239,17 @@
         <Units>milliampere</Units>
         <Count>1</Count>
     </item>
+    <!-- Temporary AuxTel only addition for CAP-423. A better long-term solution
+        would be to have an "HVBias on" configuration setting, plus code
+        which will enter fault state if the switch found to be off when
+        it should be on -->
+     <item>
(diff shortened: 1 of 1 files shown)


sal_interfaces/ATCamera/ATCamera_Telemetry.xml
```diff
@@ -239,6 +239,17 @@
...
         <Count>1</Count>
     </item>
+    <!-- Temporary AuxTel only addition for CAP-423. A better long-term solution
+        would be to have an "HVBias on" configuration setting, plus code
+        which will enter fault state if the switch found to be off when
+        it should be on -->
+     <item>
+        <EFDB_Name>hvbiasSwitch</EFDB_Name>
+        <Description>1=On, 0=Off</Description>
+        <IDL_Type>float</IDL_Type>
```
mareuter: Make this a long. Float isn't necessary.

tony-johnson: All of our CCS telemetry is floats (actually doubles), including this element. It would take special code to convert this one element to a long, and since I consider ti just a temporary workaround it is not worth the effort.

mareuter: Fair point. Looks fine otherwise.

PR #269, opened 2020-04-28
Add "raft" and "sensor" in image retrieval and oods messages

srp3rd:


mareuter:
To be clear, we should be getting 9 each of these messages for ComCam and 1 for AuxTel, yes? Main camera could potentially have up to 201 of these, yes?

srp3rd:
> To be clear, we should be getting 9 each of these messages for ComCam and 1 for AuxTel, yes? Main camera could potentially have up to 201 of these, yes?

Yes

PR #273, opened 2020-04-29
Tickets/cap 505

tony-johnson:
This change creates a new focalPlaneSummaryInfo event, with all of the information required by the header service in a format that is easy for it to use. It also adds the first subset of more generic focalPlaneSettingsApplied events (CAP-505), sufficient to test the new generic CCS->OCS configuration event processing, and to allow the remainder of the settingsApplied events to be defined in ts_xml 5.2.

### DIFF
--- a/sal_interfaces/CCCamera/CCCamera_Events.xml
+++ b/sal_interfaces/CCCamera/CCCamera_Events.xml
@@ -862,7 +862,10 @@
      settingsApplied events</Description>
     <item>
         <EFDB_Name>ccdLocation</EFDB_Name>
-        <Description>CCD location in focal-plane (e.g. R22S01) (: delimited)</Description>
+        <Description>CCD location in focal-plane (e.g. R22S01) (: delimited). 
+             This should be considered the key for all the remaining items, which will be in the same order 
+             as the locations, whether they are arrays or : delimited strings.
+        <Description>
         <IDL_Type>string</IDL_Type>
         <Units>unitless</Units>
         <Count>1</Count>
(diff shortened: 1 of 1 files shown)


sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -853,4 +853,161 @@
...
+    </item>
+    <item>
+        <EFDB_Name>ccdManufacturer</EFDB_Name>
+        <Description>Sensor manufacturer E2V or ITL (: delimited)</Description>
+        <IDL_Type>string</IDL_Type>
+        <Units>unitless</Units>
+        <Count>1</Count>
+    </item>
+    <item>
+        <EFDB_Name>ccdType</EFDB_Name>
```
mareuter, line 929: Is this necessary since we have ccdManufacturer above?

tony-johnson: There is a lot of redundant information here, but I thought the goal was that the header service would not have to apply any logic.

mareuter: Right but ccdTypes is an array of numbers not a delimited string. Is it guaranteed to be in the same order as the ccdLocations?

tony-johnson: Yes. We treat them all as arrays, and they are all in the same order, but since SAL does not support String arrays they are converted to : delimited strings.

sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -853,4 +853,161 @@
...
+        <Description>The type of each CCD</Description>
+        <IDL_Type>int</IDL_Type>
+        <IDL_Size>1</IDL_Size>
+        <Units>unitless</Units>
+        <Count>9</Count>
+        <Enumeration>E2V, ITL</Enumeration>
+    </item>
+    <item>
+        <EFDB_Name>ccdTempSetPoint</EFDB_Name>
```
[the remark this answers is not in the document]
tony-johnson: All the arrays are in the same order, and the ccdLocation should be taken as the "key" for determining which is which.

PR #276, opened 2020-05-07
Tickets/dm 24689

patrickingraham:


.gitignore
```diff
@@ -21,3 +21,4 @@ version.py
```
tribeiro, line 21: Can you remove this from your branch?

sal_interfaces/ATAOS/ATAOS_Events.xml
```diff
@@ -54,28 +54,35 @@
         <Count>1</Count>
     </item>
     <item>
-        <EFDB_Name>moveWhileExposing</EFDB_Name>
-        <Description>Move hexapod and focus while exposing enabled.</Description>
+        <EFDB_Name>atspectrograph</EFDB_Name>
```
tribeiro, line 57: Are you removing the `moveWhileExposing` item of the event?

sal_interfaces/ATAOS/ATAOS_Events.xml
```diff
@@ -54,28 +54,35 @@
...
-        <IDL_Type>float</IDL_Type>
-        <Units>millimeter</Units>
+        <EFDB_Name>moveWhileExposing</EFDB_Name>
+        <Description>Move hexapod and focus while exposing enabled.</Description>
+        <IDL_Type>boolean</IDL_Type>
+        <Units>unitless</Units>
         <Count>1</Count>
     </item>
 </SALEvent>
+<!--<SALEvent>-->
```
tribeiro, line 71: maybe remove the comment lines?

sal_interfaces/ATAOS/ATAOS_Events.xml
```diff
@@ -442,4 +449,47 @@
...
 </SALEvent>
+<SALEvent>
+    <Subsystem>ATAOS</Subsystem>
+    <Version>3.8</Version>
+    <Author>Tiago Ribeiro</Author>
+    <EFDB_Topic>ATAOS_logevent_focusOffsetSummary</EFDB_Topic>
+    <Alias>focusOffsetSummary</Alias>
+    <Explanation>http://sal.lsst.org</Explanation>
+    <item>
+        <EFDB_Name>totalFocusCorrectionOffset</EFDB_Name>
```
tribeiro, line 460: So, the name of the topic is already `focusOffsetSummary`, I think the attributes could have simpler names. I would suggest:

`totalFocusCorrectionOffset` -> `total`
`cumulativeUserOffset` -> `user`
`filterOffset` -> `filter`
`disperserOffset` -> `disperser`
`wavelengthOffset` -> `wavelength`

patrickingraham: renaming completed. Note that this affected both the XML as well as the test_csc and ataos_csc files, which are being modified under the same ticket.

patrickingraham:
made all changed *except* the change history, due to the question above.

tribeiro:
> Will definitely remove the .gitignore and implement a global one. However, if the commit history will get squashed before the merge, do I have to fix the commit history?

So, squash-and merge is not a good option as it messes up with the main branch history. It looks like you committed to the branch directly instead or merging a branch.

PR #291, opened 2020-06-10
Tickets/dm 25352

teweitsai:


sal_interfaces/MTM2/MTM2_Commands.xml
```diff
@@ -101,4 +101,46 @@
...
+        <Description>0: turn off FB. 1: turn on FB.</Description>
+        <IDL_Type>int</IDL_Type>
+        <Units>unitless</Units>
+        <Count>1</Count>
+    </item>
+</SALCommand>
+
+<SALCommand>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_command_useZenithAngleFromMount</EFDB_Topic>
```
tribeiro, line 120: So, the mount publishes Elevation and not Zenit Angle. But maybe it would be a better idea to make this more generic. Also, how would we switch back to using the on-board instruments? Based on  the payload this is just a "on" command. I would suggest we use a command like:

`selectInclinationTelemetrySource` and the payload would be `source` which can be an enumeration with the options like `ONBOARD` and `MOUNT`.

sal_interfaces/MTM2/MTM2_Commands.xml
```diff
@@ -101,4 +101,46 @@
...
+    </item>
+</SALCommand>
+
+<SALCommand>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_command_setTemperatureOrigin</EFDB_Topic>
+    <Alias>setTemperatureOrigin</Alias>
+    <Description>Set the origin (offset) of temperature sensors.</Description>
+    <item>
+        <EFDB_Name>origin</EFDB_Name>
```
tribeiro, line 138: I think this attribute would be better named as `temperature` instead of `origin`...

sal_interfaces/MTM2/MTM2_Commands.xml
```diff
@@ -101,4 +101,46 @@
...
 
+<SALCommand>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_command_switchForceBalanceSystem</EFDB_Topic>
+    <Alias>switchForceBalanceSystem</Alias>
+    <Description>Switch the force balance (FB) system.</Description>
+    <item>
+        <EFDB_Name>status</EFDB_Name>
+        <Description>0: turn off FB. 1: turn on FB.</Description>
+        <IDL_Type>int</IDL_Type>
```
tribeiro, line 112: If it is on/off, I would recommend using a `boolean`.

sal_interfaces/MTM2/MTM2_Events.xml
```diff
@@ -101,4 +101,32 @@
     </item>
 </SALEvent>
 
+<SALEvent>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_logevent_isForceBalanceSystemOn</EFDB_Topic>
+    <Alias>isForceBalanceSystemOn</Alias>
+    <Description>Force balance (FB) system is on or off?</Description>
+    <item>
+        <EFDB_Name>status</EFDB_Name>
```
tribeiro, line 110: What about just naming this `on`? and it could be True/False?

sal_interfaces/MTM2/MTM2_Events.xml
```diff
@@ -101,4 +101,32 @@
     </item>
 </SALEvent>
 
+<SALEvent>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_logevent_isForceBalanceSystemOn</EFDB_Topic>
```
tribeiro, line 106: What about `forceBalanceSystemStatus`?

sal_interfaces/MTM2/MTM2_Events.xml
```diff
@@ -101,4 +101,32 @@
...
+        <Description>Ture if FB is on, False if off.</Description>
+        <IDL_Type>boolean</IDL_Type>
+        <Units>unitless</Units>
+        <Count>1</Count>
+    </item>
+</SALEvent>
+
+<SALEvent>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_logevent_isZenithAngleFromMount</EFDB_Topic>
```
tribeiro, line 120: If you accept my proposal to rename the command to `selectInclinationTelemetrySource` I would recommend renaming this `inclinationTelemetrySource` and the payload would be changed to be the enumeration value instead of `isZenithAngleFromMount`.

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -51,6 +51,54 @@
     </item>
 </SALTelemetry>
 
+<SALTelemetry>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_mirrorPositionMeasuredIMS</EFDB_Topic>
```
tribeiro, line 56: I suggest renaming this topic to `positionIMS`.

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -136,28 +184,35 @@
 <SALTelemetry>
     <Subsystem>MTM2</Subsystem>
     <EFDB_Topic>MTM2_temperaturesMeasured</EFDB_Topic>
```
tribeiro, line 186: Consider renaming this to `temperature` only...

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -272,6 +327,54 @@
     </item>
 </SALTelemetry>
 
+<SALTelemetry>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_balanceForceData</EFDB_Topic>
```
tribeiro, line 332: I suggest `forceBalance`... Always put what you are measuring first.  :)

sal_interfaces/MTM2/MTM2_Events.xml
bxin: I like inclinationTelemetrySource better

sal_interfaces/MTM2/MTM2_Telemetry.xml
bxin: I think MTM2_mirrorPositionMeasuredIMS is fine, since we already have MTM2_mirrorPositionMeasured

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -136,28 +184,35 @@
...
         <EFDB_Name>exhaustTemperature</EFDB_Name>
         <Description>Exhaust temperatures: #1 and #2.</Description>
         <IDL_Type>double</IDL_Type>
-        <Units>Celsius</Units>
+        <Units>deg_C</Units>
         <Count>2</Count>
     </item>
+    <item>
+        <EFDB_Name>origin</EFDB_Name>
+        <Description>Origin (offset).</Description>
```
bxin, line 211: Shouldn't this be in an event, which gets published after we set the offset using the command?

sal_interfaces/MTM2/MTM2_Telemetry.xml
bxin: I think this refers to the `force balance system`. I am OK either way.

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -272,6 +327,54 @@
     </item>
 </SALTelemetry>
 
+<SALTelemetry>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_balanceForceData</EFDB_Topic>
+    <Description>Net forces and moments as commanded by the force balance system. This is a component of the total force, which has the topics of netForcesTotal and netMomentsTotal.</Description>
```
bxin, line 333: suggest removing `which has the topics of netForcesTotal and netMomentsTotal`. It just adds confusion.

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -272,6 +327,54 @@
...
+    <item>
+        <EFDB_Name>fx</EFDB_Name>
+        <Description>Net force in x direction.</Description>
```
bxin, line 336: suggest `total x-force commanded by the force balance system`. Same for the 5 other parameters below.

sal_interfaces/MTM2/MTM2_Commands.xml
teweitsai: Based on the document of M2 wish list, there should be only one offset value to apply to all temperature sensors. Can you help to check your document and decide you want single offset value or 16 values?

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: This name is based on the LTS-162 v5.

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: This is based on the LTS-162 v5.

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: This is based on the LTS-162 v5.

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: I also considered to use the event in the initial beginning. However, I felt it might be easier to do the calculation if we put the origin/ offset and temperature together if the analyzer does not know he/she should check the event to get the origin value. I can put it as an event if you like. Let us discuss this.

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: Maybe you could suggest a good description for this. I felt if I do not add this, it confused me as well. I do not understand this topic when I read the LTS-162. This is why I slacked you in the last week to understand it.

sal_interfaces/MTM2/MTM2_Commands.xml
teweitsai: If we use the name of `MTM2_command_setTemperatureOffset`, the attribute name is better to use `offset`. However, it is a reserved word that can not be used.

sal_interfaces/MTM2/MTM2_Commands.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Commands.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Events.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Events.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Events.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: Done in commit: 9c5cc3f. Thanks!

sal_interfaces/MTM2/MTM2_Telemetry.xml
bxin: Net forces and moments as commanded by the force balance system. These are calculated using the 72 axial and 6 tangent balance forces, which are a component of the total forces being commanded.

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: Updated in commit: 7f353f2. Thanks!

sal_interfaces/MTM2/MTM2_Telemetry.xml
teweitsai: Updated in commit: 7f353f2. Thanks!

sal_interfaces/MTM2/MTM2_Commands.xml
teweitsai: Updated in commit: 7f353f2. Thanks!

bxin: use ringOffset, intakeOffset, exhaustOffset for attributes

sal_interfaces/MTM2/MTM2_Telemetry.xml
bxin: discussed with Te-Wei. We will make this an event.

sal_interfaces/MTM2/MTM2_Events.xml
```diff
@@ -101,4 +101,60 @@
...
 </SALEvent>
 
+<SALEvent>
+    <Subsystem>MTM2</Subsystem>
+    <EFDB_Topic>MTM2_logevent_forceBalanceSystemStatus</EFDB_Topic>
+    <Alias>forceBalanceSystemStatus</Alias>
+    <Description>The status of force balance (FB) system (on or off).</Description>
+    <item>
+        <EFDB_Name>status</EFDB_Name>
+        <Description>Ture if FB is on, False if off.</Description>
```
tribeiro, line 111: typo `Ture` -> `True`

sal_interfaces/MTM2/MTM2_Telemetry.xml
tribeiro: So, I think it is already implicit that this is a "mirror" position because the component is `M2`... I'm thinking in terms of someone looking at the interface and trying to build a query for it... it is much simpler and easy to understand `positionIMS` then  `mirrorPositionMeasuredIMS`... 

Most of my further comments are on those lines as well...

sal_interfaces/MTM2/MTM2_Telemetry.xml
```diff
@@ -136,26 +184,26 @@
 <SALTelemetry>
     <Subsystem>MTM2</Subsystem>
     <EFDB_Topic>MTM2_temperaturesMeasured</EFDB_Topic>
-    <Description>Mirror and cell temperature sensor readings.</Description>
+    <Description>Mirror and cell temperature sensor readings relative to the origin (offset).</Description>
     <item>
         <EFDB_Name>ringTemperature</EFDB_Name>
```
tribeiro, line 189: I just noticed that the attributes also have "temperature" on them.. I would recommend removing "Temperature" from the topic attributes because we already know they are temperatures from the topic name. 

So, instead of `ringTemperature` just `ring` would be fine, and so on...

sal_interfaces/MTM2/MTM2_Events.xml
teweitsai: Done in commit: 3806fac. Thanks!

PR #292, opened 2020-06-11
DM-25339: Update Dome XML

r-owen:


sal_interfaces/Dome/Dome_Commands.xml
```diff
@@ -1,241 +1,133 @@
...
+            <Description>The velocity. Positive is the direction of increasing azimuth.</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>deg/second</Units>
+            <Count>1</Count>
+        </item>
+    </SALCommand>
+    <SALCommand>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_command_moveEl</EFDB_Topic>
+        <Description>Move the elevation axis (light/wind screen) at constant velocity, starting from a specified position.</Description>
```
wvreeven, line 50: The moveEl command only accepts a location and no velocity so this description is incorrect.

sal_interfaces/Dome/Dome_Commands.xml
```diff
@@ -1,241 +1,133 @@
...
+            <Units>deg/second</Units>
+            <Count>1</Count>
+        </item>
+    </SALCommand>
+    <SALCommand>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_command_moveAz</EFDB_Topic>
+        <Description>Move the azimuth axis at constant velocity, starting from a specified position.</Description>
+        <item>
+            <EFDB_Name>azimuth</EFDB_Name>
```
wvreeven, line 33: Didn't we agree on slack to rename this to position?

sal_interfaces/Dome/Dome_Commands.xml
```diff
@@ -1,241 +1,133 @@
...
+            <Units>deg/second</Units>
+            <Count>1</Count>
+        </item>
+    </SALCommand>
+    <SALCommand>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_command_moveEl</EFDB_Topic>
+        <Description>Move the elevation axis (light/wind screen) at constant velocity, starting from a specified position.</Description>
+        <item>
+            <EFDB_Name>elevation</EFDB_Name>
```
wvreeven, line 52: Again, shouldn't this have been renamed to position?

sal_interfaces/Dome/Dome_Commands.xml
```diff
@@ -1,241 +1,133 @@
...
+            <Description>The position.</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>deg</Units>
+            <Count>1</Count>
+        </item>
+    </SALCommand>
+    <SALCommand>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_command_park</EFDB_Topic>
```
wvreeven, line 61: Sorry to be picky because I really appreciate that you're making all these changes but a Description would be welcome here.

sal_interfaces/Dome/Dome_Commands.xml
```diff
@@ -1,241 +1,133 @@
...
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_command_setLouvers</EFDB_Topic>
+        <Description>Move one or more louvers.</Description>
+        <item>
+            <EFDB_Name>louvers</EFDB_Name>
+            <Description>Bitmask of louvers to move.</Description>
+            <IDL_Type>int</IDL_Type>
+            <Units>unitless</Units>
+            <Count>1</Count>
+        </item>
```
wvreeven, line 73: Interesting change from ID to bitmask. Should this refer to an enum like in the events file?

sal_interfaces/Dome/Dome_Events.xml
```diff
@@ -1,268 +1,157 @@
...
-                <Count>1</Count>
-            </item>
-</SALEvent>
+    <SALEvent>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_logevent_brakesEngaged</EFDB_Topic>
+        <item>
+            <EFDB_Name>brakes</EFDB_Name>
+            <Description>Bitmask of brakes that are engaged.
+            The bits are defined by the Brake enum.</Description>
```
wvreeven, line 127: Where is this enum defined?

sal_interfaces/Dome/Dome_Events.xml
```diff
@@ -1,268 +1,157 @@
...
+            <Count>1</Count>
+        </item>
+    </SALEvent>
+    <SALEvent>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_logevent_lockingPinsEngaged</EFDB_Topic>
+        <item>
+            <EFDB_Name>engaged</EFDB_Name>
+            <Description>Bitmask of locking pins that have been engaged.
+            The bits are defined by the LockingPin enum.</Description>
```
wvreeven, line 139: Where is this enum defined?

sal_interfaces/Dome/Dome_Events.xml
```diff
@@ -1,268 +1,157 @@
...
+            <Count>1</Count>
+        </item>
+    </SALEvent>
+    <SALEvent>
+        <Subsystem>Dome</Subsystem>
+        <EFDB_Topic>Dome_logevent_interlocks</EFDB_Topic>
+        <item>
+            <EFDB_Name>interlocks</EFDB_Name>
+            <Description>Bitmask of safety interlocks that are latched.
+            The bits are defined by the Interlock enum.</Description>
```
wvreeven, line 151: Where is this enum defined?

sal_interfaces/Dome/Dome_Telemetry.xml
```diff
@@ -1,397 +1,389 @@
...
+        <Description>Status of the azimuth axis and drives.</Description>
+        <item>
+            <EFDB_Name>positionActual</EFDB_Name>
+            <Description>Measured azimuth axis position.</Description>
+            <IDL_Type>float</IDL_Type>
+            <Units>deg</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>positionCommand</EFDB_Name>
```
wvreeven, line 16: Should this be positionCommanded? Same for all other occurrences of Command in the rest of this file.

sal_interfaces/Dome/Dome_Commands.xml
r-owen: Great!

sal_interfaces/Dome/Dome_Commands.xml
r-owen: I agree, but I'd have to know what it does first! I am guessing it moves the azimuth to a given "park" position, but what about elevation, the shutter, etc.?

sal_interfaces/Dome/Dome_Commands.xml
r-owen: Since there is a crawlEl command I am guessing that this command should have a velocity as well. I assume EIE will use the same basic code to run both axes. What do you think?

sal_interfaces/Dome/Dome_Commands.xml
r-owen: Yes. Which is why I asked you for the names of the louvers. I wanted to make the bitmask.

sal_interfaces/Dome/Dome_Events.xml
r-owen: It will be defined in the Events file once I know the names of the brakes. If they only have numbers then are they numbered from 0 or 1?

sal_interfaces/Dome/Dome_Events.xml
r-owen: Once I know the names of the locking pins it will be defined in this file.

sal_interfaces/Dome/Dome_Telemetry.xml
r-owen: OK. I'll make that change.

sal_interfaces/Dome/Dome_Events.xml
r-owen: Again I'd need the names of the interlocks to make the bitmask.

I guess the question is whether to start with stupid names and then improve them later or just wait until we know more. I lean towards the latter.

sal_interfaces/Dome/Dome_Commands.xml
wvreeven: That is a good point which we haven't discussed yet. For now the mock controller assumes that only the dome azimuth rotation gets parked and the rest not.

sal_interfaces/Dome/Dome_Commands.xml
wvreeven: My assumption was to do the translation from numeric IDs to the louvers matrix IDs inside the Dome CSC. However, I have sent the EIE documentation regarding the matrix IDs to you.

sal_interfaces/Dome/Dome_Events.xml
wvreeven: We haven't discussed this yet. My assumption was that the brakes are all engaged or none are. So let's clarify this and then make an update to this event.

sal_interfaces/Dome/Dome_Events.xml
wvreeven: We haven't discussed this yet. My assumption was that the locking pins are all engaged or none are. So let's clarify this and then make an update to this event.

sal_interfaces/Dome/Dome_Events.xml
wvreeven: We haven't discussed this yet. So let's clarify this and then make an update to this event.

PR #294, opened 2020-06-11
Fix target event field types.

r-owen:


sal_interfaces/Dome/Dome_Events.xml
```diff
@@ -104,15 +104,15 @@
...
-            <Description>Target position; nan if crawling.</Description>
-            <IDL_Type>int</IDL_Type>
+            <Description>Target position; nan for the crawlEl command.</Description>
+            <IDL_Type>float</IDL_Type>
             <Units>deg</Units>
             <Count>1</Count>
         </item>
         <item>
             <EFDB_Name>velocity</EFDB_Name>
             <Description>Target velocity.</Description>
```
wvreeven, line 114: The velocity should be nan for the moveEl command since it will move to a position and then stop.

r-owen: Actually it should be 0 because the velocity will be known to be 0 (whereas target position truly makes no sense for the crawlEl command). I expanded the description.

r-owen:
I also added a Louver enum (to Dome_Event.xml, which is where we usually put enumerations) that gives a name to each louver index. I specified explicit values so the first one could be 0 (instead of the default of 1), since C++ and Python both use 0-based indices.

I intentionally declined to reference Dome_Events.xml in the text that mentions the enumeration because:
* That is our standard location
* The 3 files are combined in rendered documentation (which is what I hope most folks will read) and in ts_idl, so the location is not relevant in most cases.

PR #295, opened 2020-06-12
Fix Dome enumeration.

tribeiro:
Add unit test to check enumeration integrity.

### DIFF
--- a/tests/test_enumeration.py
+++ b/tests/test_enumeration.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import re
+import pytest
+import xml.etree.ElementTree as et
+import lsst.ts.xml as ts_xml
+
+
+@pytest.mark.parametrize("xmlfile,csc,topic", ts_xml.get_xmlfile_csc_topic())
+def test_enumeration(xmlfile, csc, topic):
+    """Test that enumerations have the correct format.
+
+    Parameters
+    ----------
+    xmlfile : `pathlib.Path`
+        Full filepath to the Events XML file for the CSC.
+    csc : `testutils.subsystems`
+        Name of the CSC
+    """
+    # Test the topic <EFDB_Name> field.
+    with open(str(xmlfile), "r", encoding="utf-8") as f:
+        tree = et.parse(f)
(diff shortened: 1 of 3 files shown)


tests/test_enumeration.py
```diff
@@ -0,0 +1,26 @@
...
+        Name of the CSC
+    """
+    # Test the topic <EFDB_Name> field.
+    with open(str(xmlfile), "r", encoding="utf-8") as f:
+        tree = et.parse(f)
+    root = tree.getroot()
+
+    for sal_enum in root.findall("Enumeration"):
+        for enum_value in sal_enum.text.split(','):
+            assert len(enum_value.rstrip().lstrip()) > 0, f"Invalid enumeration value in {csc}/{topic}."
```
r-owen, line 26: Consider `enum_value.strip()` instead of rstrip followed by lstrip.

I strongly suggest using a regex to test the stripped values, probably: `[A-Z][a-zA-Z0-9_]*`.

tribeiro:
I did the regex check. Not sure if the values of the enumerations should be checked though. I thought enumerations could have float values as well.

tests/test_enumeration.py
```diff
@@ -0,0 +1,31 @@
...
+        Full filepath to the Events XML file for the CSC.
+    csc : `testutils.subsystems`
+        Name of the CSC
+    """
+    # Test the topic <EFDB_Name> field.
+    with open(str(xmlfile), "r", encoding="utf-8") as f:
+        tree = et.parse(f)
+    root = tree.getroot()
+
+    reg_exp = re.compile("^[A-Z][a-zA-Z0-9_]\\s?=?\\s?(\\d)?")
```
r-owen, line 25: Please consider the following because it is pickier about the values. It works for all the cases I could find. It insists that hex values start with 0x or 0X, which I think is reasonable. The leading "r" means you don't have to double the backslashes.
```
r"^[A-Z][a-zA-Z0-9_]*(\s*=\s*((0[xX][0-9a-fA-F]+)|(\d+)))?$"
```

r-owen:
For the record: SAL rejects float and string enum values (I just checked -- including strings with double quotes and without).

PR #299, opened 2020-06-16
Fix degree to deg_c(Celsius) in TunableLaser

couger01:


sal_interfaces/TunableLaser/TunableLaser_Telemetry.xml
```diff
@@ -10,49 +10,49 @@
             <EFDB_Name>tk6_temperature</EFDB_Name>
             <Description>The temperature sensors for the laser.</Description>
             <IDL_Type>float</IDL_Type>
-            <Units>degree</Units>
+            <Units>Celsius</Units>
```
pkubanek, line 13: please use deg_C. Thanks

wvreeven:
I had a discussion about this with Russell and aliases are acceptable as well in this project, so Celsius is fine.

tribeiro:
I hate to drag this too long but I'd rather we all adopt the same naming convention. I know `Celsius` is fine and we allow alias, but I'd rather have everyone using the recommended (e.g. `deg_C`).

PR #309, opened 2020-07-02
Added timestamp to all ESS telemetry topics.

wvreeven:


sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -10,7 +10,7 @@
       <Description>Air Turbulence Instrument - CSAT3B (Tri-Axial) Serial to Ethernet Converter</Description>
       <item>
           <EFDB_Name>ux</EFDB_Name>
-	  <Description>x-axis wind speed</Description>
+	      <Description>x-axis wind speed</Description>
```
pkubanek, line 13: Looks like you/your editor added a tab. Remove it please (also in following <Description> changes).

sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -255,12 +300,21 @@
...
           <EFDB_Name>rainLevel</EFDB_Name>
-	  <Description>Rain Level Telemetry</Description>
+	      <Description>Rain Level Telemetry</Description>
           <IDL_Type>float</IDL_Type>
           <Units>mm</Units>
           <Conversion></Conversion>
           <Sensor_location></Sensor_location>
           <Count>1</Count>
       </item>
+      <item>
```
pkubanek, line 310: Please move that as a first item. Seems as consensus among subsystems I know (M1M3,MTMount) - timestamp is the key and therefore the most important (kind of) part of the message, and should be on top.

sal_interfaces/ESS/ESS_Telemetry.xml
wvreeven: Good catch. I think it was already in there but nevertheless the file shouldn't contain tabs. I fixed it.

sal_interfaces/ESS/ESS_Telemetry.xml
wvreeven: Done.

wvreeven:
I addressed both of your points. Can you review again please?

PR #310, opened 2020-07-03
Fix doc warnings

couger01:


doc/conf.py
```diff
@@ -20,5 +20,5 @@
 except KeyError:
     pass
 
-extensions.append('sphinx.ext.autosectionlabel')
-autosectionlabel_prefix_document=True
\ No newline at end of file
+# extensions.append('sphinx.ext.autosectionlabel')
+# autosectionlabel_prefix_document=True
```
r-owen, line 24: What does this change (commenting out these two lines) do?

doc/generate_cet.py
```diff
@@ -36,8 +36,11 @@ def main():
                     cf.write("\n")
                     continue
 
+                
                 topic_name = topic.find('EFDB_Topic').text
                 short_name = topic_name.split("_")[-1]
+                # short_name = '_'.join(split_short_name)
```
r-owen, line 42: It would be safer to specify the known number of splits (2 for commands and events, 1 for telemetry) in your call to `split`. That way a non-compliant name with an underscore would be handled correctly.

Please don't commented-out code (especially with no explanation)

doc/sal_constraints_and_recommendations.rst
```diff
@@ -561,6 +595,8 @@ An easy way to tell if a particular unit is valid is to try to make an astropy.u
     These units may not make sense but are valid according to the astropy parser.
     Please note this "feature" of astropy.
```
r-owen, line 596: It would be helpful to say that these units are compatible with Astropy.

And I strongly suggest replacing this current comment with that information. The current comment seems needlessly snarky. What "feature" are you referring to?

doc/conf.py
r-owen: Please do not leave commented-out code. It leads to bit rot. If you absolutely must for some reason then provide a code comment explaining what it's all about. But far better to just delete it.

couger01: Adds a autogenerated label to each section and adds a prefix based on the name of the file to that label, but it doesn't create unique enough labels leading to all of those previous warnings. 1658 down to 19.

doc/sal_constraints_and_recommendations.rst
couger01: This comment, which I didn't write, refers to astropy excepting units such "deg22" which apparently is a degree in the 22nd power. Apparently it is a nonsensical unit based on the comment and so people should be wary of the feature according to the comment.

doc/generate_cet.py
couger01: It works, thanks Russell. Removed commented out code.

doc/conf.py
couger01: Removed it.

doc/sal_constraints_and_recommendations.rst
r-owen: I would appreciate it.

doc/sal_constraints_and_recommendations.rst
```diff
@@ -372,6 +380,8 @@ The *<Subsystem>* and *<Alias>* tags for command's and logevent's must be consis
 
 ALL names must be less than 64 characters in length.
 
+.. _sal_constraints_and_recommendations:timestamps:
```
r-owen, line 383: At line 350 above "These should not have any embedded no alphanumeric characters" please fix no->non. Better yet reword that bit and the next one to just state that the only characters allowed are a-zA-Z0-9_ and that the name must start with a-z. I think it would be much clearer than trying to list all prohibited characters.

r-owen: Also the section "Ignored Attributes in Topics" is obsolete. If you are willing, please delete that section.

SAL 4.1 added the ability to define topics with no user-defined fields (and we are using 4.1.4, so we no longer need this old rule). Ignored fields are now deprecated.

doc/sal_constraints_and_recommendations.rst
```diff
@@ -383,6 +393,8 @@ If a time-of-data is to be associated with an item it should be named
 
 The time(s) should be obtained using the SAL getCurrentTime() method, which returns a double precision value of TAI time with a resolution of at least 0.001 seconds.
 
+.. _sal_constraints_and_recommendations:ignored-attributes-in-topics:
```
r-owen, line 396: In the paragraph "The time(s)..." please consider "value of TAI time in unix seconds, with a resolution..." (add "in unix seconds")

doc/sal_constraints_and_recommendations.rst
```diff
@@ -509,9 +537,13 @@ Alternatively, upload the *testingest.sql* using the TBD webpage.
 
 `Proposed method of recording subsystem "Settings" data <https://confluence.lsstcorp.org/pages/viewpage.action?pageId=58949768>`_
 
+.. _sal_constraints_and_recommendations:sal-topic-and-parameter-<description>-and-<unit>-fields:
```
r-owen, line 540: Please don't use < and > in section labels. Even if sphinx can handle it without complaining, it makes for needlessly awkward URLs.

If you are willing to change this, I suggest a global search for it -- I've found at least three instances.

doc/sal_constraints_and_recommendations.rst
```diff
@@ -381,23 +393,31 @@ If a time-of-data is to be associated with an item it should be named
...
 
-The time(s) should be obtained using the SAL getCurrentTime() method, which returns a double precision value of TAI time with a resolution of at least 0.001 seconds.
+The time(s) should be obtained using the SAL getCurrentTime() method, which returns a double precision value of TAI time in unix seconds with a resolution of at least 0.001 seconds.
+
+.. _sal_constraints_and_recommendations:ignored-attributes-in-topics:
 
 Ignored Attributes in Topics
 ============================
 
+*Deprecated in SAL 4.1*
```
r-owen, line 403: Thank you. Please consider appending a brief explanation of why it is deprecated, e.g. "because topics with no fields are allowed"

PR #311, opened 2020-07-06
adding file path to MTMount simiulator software

andrewbheyer:
Because some simulator software does not exist on their gitlap repos I compiled all the simulators into a zip file in the La Serena PDM server.

### DIFF
--- a/sal_interfaces/SALSubsystems.xml
+++ b/sal_interfaces/SALSubsystems.xml
@@ -93,7 +93,7 @@
   <SoftwareLanguage>Labview</SoftwareLanguage>
   <RuntimeLanguages>IDL</RuntimeLanguages>
   <VendorContact>Tekniker</VendorContact>
-  <Simulator>No</Simulator>
+  <Simulator>LaSerenaPDM/TSS-Share/TMA</Simulator>
   <Configuration>Not Configurable</Configuration>
 </SALSubsystem>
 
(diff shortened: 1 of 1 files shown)


sal_interfaces/SALSubsystems.xml
```diff
@@ -93,7 +93,7 @@
   <SoftwareLanguage>Labview</SoftwareLanguage>
   <RuntimeLanguages>IDL</RuntimeLanguages>
   <VendorContact>Tekniker</VendorContact>
-  <Simulator>No</Simulator>
+  <Simulator>LaSerenaPDM/TSS-Share/TMA</Simulator>
```
r-owen, line 96: Can you make this a URL? Here's the start of the text for this tag from https://confluence.lsstcorp.org/display/LTS/SALSubsystem+Attributes:

This contains the Github repository URL that contains the simulator code. If the repository is privately held, the URL should be appended with a space and (private).

andrewbheyer: Agreed, I will make these changes within the next few days. I am moving the tstn documentation into it's own repository for now. This decision came out of the documentation meetings - they don't want the TMA documentation to be inside of a technote.

Once I have this moved I will update the link to point to this new location.

andrewbheyer:
I have made a new repository to host the tma simulation software. This will hold the basic simulators along with the hardware simulators (most notably the speedgoat which is a rigid body simulation of the alt az motion with 1ms response). The tstn-008 technote will be deleted.

andrewbheyer:
Suggestion read, the README is created and the repository is no longer empty. You may see some differences to the documentation you see there and the documentation you saw on tstn-008. This is because I did 2 main fixes to make that I think will make the documentation better."

1) Removed redundant use of "TMA" from titles and titles names
2) Created a "Credentials" page to host the information for accessing the PDM server and the nexus server.

PR #319, opened 2020-07-09
Consolidate generic topics into CSC RST documentation

couger01:
* Fixed side bar in CSC to only show topic names
* Placed enumerations into own section


doc/generate_cet.py
```diff
@@ -3,7 +3,121 @@
...
+def add_generics(cf, subsystem, set_name, has_generics, has_specific):
+    ignored_attributes, ignored_fields = provide_ignored()
+    if has_generics is True:
+        gen_tree = ET.parse("../sal_interfaces/SALGenerics.xml")
+        gen_root = gen_tree.getroot()
+        for gen_set in gen_root:
+            gen_set[:] = sorted(gen_set, key=lambda child: (child.tag, child.find("EFDB_Topic").text.split("_")[-1] if child.find("EFDB_Topic") is not None else child.tag))
+            gen_set_name = gen_set.tag
+            if gen_set_name == set_name:
+                if gen_set_name[3:-3] not in has_specific:
```
r-owen, line 20: What is the `[3:-3]` for? A code comment would be helpful.

doc/generate_cet.py
```diff
@@ -3,7 +3,121 @@
...
+    ignored_fields = ["Description", "Alias", "Device", "Property", "Action", "Value", "Subsystem", "Version", "Author", "Explanation"]
+    return ignored_attributes, ignored_fields
+
+def add_generics(cf, subsystem, set_name, has_generics, has_specific):
+    ignored_attributes, ignored_fields = provide_ignored()
+    if has_generics is True:
+        gen_tree = ET.parse("../sal_interfaces/SALGenerics.xml")
+        gen_root = gen_tree.getroot()
+        for gen_set in gen_root:
+            gen_set[:] = sorted(gen_set, key=lambda child: (child.tag, child.find("EFDB_Topic").text.split("_")[-1] if child.find("EFDB_Topic") is not None else child.tag))
```
r-owen, line 17: I'm puzzled why you sort the topics here yet somehow manage to interleave generic and non-generic topics in alphabetical order. A code comment explaining what's going on would help future maintainers of the code.

doc/generate_cet.py
```diff
@@ -3,7 +3,121 @@
 import pathlib
 
 
+def provide_ignored():
+    ignored_attributes = ["EFDB_Name", "Description"]
+    ignored_fields = ["Description", "Alias", "Device", "Property", "Action", "Value", "Subsystem", "Version", "Author", "Explanation"]
+    return ignored_attributes, ignored_fields
+
+def add_generics(cf, subsystem, set_name, has_generics, has_specific):
```
r-owen, line 11: Please add a doc string explaining what this does and what the arguments are.

doc/generate_cet.py
```diff
@@ -3,7 +3,121 @@
 import pathlib
 
 
+def provide_ignored():
```
r-owen, line 6: Please add a doc string to this and all other functions. It makes the code much easier to understand and maintain.

doc/generate_cet.py
```diff
@@ -3,7 +3,121 @@
...
+    ignored_attributes, ignored_fields = provide_ignored()
+    if has_generics is True:
+        gen_tree = ET.parse("../sal_interfaces/SALGenerics.xml")
+        gen_root = gen_tree.getroot()
+        for gen_set in gen_root:
+            gen_set[:] = sorted(gen_set, key=lambda child: (child.tag, child.find("EFDB_Topic").text.split("_")[-1] if child.find("EFDB_Topic") is not None else child.tag))
+            gen_set_name = gen_set.tag
+            if gen_set_name == set_name:
+                if gen_set_name[3:-3] not in has_specific:
+                    cf.write(f"{gen_set_name[3:-3]}s\n")
```
r-owen, line 21: Also where does the variable `cf` come from? You write to it, but it is not in the argument list. I very strongly suggest passing all values as arguments (or using a class and putting the value in "self"). Global variables make code much harder to maintain.

doc/generate_cet.py
```diff
@@ -15,6 +129,7 @@ def main():
     f.write(".. toctree::\n  :glob:\n  :maxdepth: 1\n\n  *\n\n")
     subsystem_tree = ET.parse("../sal_interfaces/SALSubsystems.xml")
     for subsystem in testutils.subsystems:
+        has_specific = []
```
r-owen, line 132: What is has_specifics? The name makes it sound like a list of boolean. A code comment would help. If it is not a list of boolean then please consider a clearer name.

doc/generate_cet.py
```diff
@@ -3,7 +3,144 @@
...
+    cf : `IO`
+        A file object for writting to each CSC rst file.
+    subsystem : `str`
+        The name of the CSC.
+    set_name : `str`
+        The name of the topic set.
+    has_generics : `bool` or `List` of `str`
+        It either has generic topics(yes) or not(no) or a list of supported topics.
+    has_specific : `List` of `str`
+        A list for specifying whether a CSC has specific kinds of topics or not(command, event, telemetry)
```
r-owen, line 26: What is in the list? Three booleans? The names of the kinds? Does case matter? I am guessing it is a list of one or more of "command", "event", "telemetry" but I honestly don't know.

doc/generate_cet.py
```diff
@@ -3,7 +3,144 @@
...
+    Parameters
+    ----------
+    cf : `IO`
+        A file object for writting to each CSC rst file.
+    subsystem : `str`
+        The name of the CSC.
+    set_name : `str`
+        The name of the topic set.
+    has_generics : `bool` or `List` of `str`
+        It either has generic topics(yes) or not(no) or a list of supported topics.
```
r-owen, line 24: Please consider something along the lines of "True if it has all generic topics, False if it has no generic topics, a list of generic topic names if it has some generic topics."

doc/generate_cet.py
r-owen: Still want a code comment

doc/generate_cet.py
```diff
@@ -3,7 +3,144 @@
...
+    """
+    ignored_attributes, ignored_fields = provide_ignored()
+    if has_generics is True:
+        gen_tree = ET.parse("../sal_interfaces/SALGenerics.xml")
+        gen_root = gen_tree.getroot()
+        for gen_set in gen_root:
+            gen_set[:] = sorted(gen_set, key=lambda child: (child.tag, child.find("EFDB_Topic").text.split("_")[-1] if child.find("EFDB_Topic") is not None else child.tag))
+            gen_set_name = gen_set.tag
+            if gen_set_name == set_name:
+                if gen_set_name[3:-3] not in has_specific: # Remove SAL and Set from the string in order to compare topic set name correctly
```
r-owen, line 37: Very helpful. Unfortunately, it extends well pas column 79; I suggest moving the comment to just above the "if" statement and wrapping it as needed.

doc/generate_cet.py
r-owen: Still want a code comment.

doc/generate_cet.py
```diff
@@ -41,20 +180,37 @@ def main():
...
-            cf.write(f"{dds_type}\n")
-            cf.write(f"{'-'*len(dds_type)}\n")
+                add_generics(cf, subsystem, set_name=f"SAL{dds_type[:-1]}Set", has_generics=has_generics, has_specific=has_specific)
+                continue
+            if dds_type != "Events":
+                cf.write(f"{dds_type}\n")
+                cf.write(f"{'-'*len(dds_type)}\n")
             root = tree.getroot()
+            set_name = root.tag
+            # root[:] = sorted(root, key=lambda child: child.get("EFDB_Topic").split("_")[-1] if child.get("EFDB_Topic") is not None else child.tag)
```
r-owen, line 192: Please delete comment-out code.

doc/generate_cet.py
```diff
@@ -3,7 +3,144 @@
...
+                            cf.write("\n")
+                    else:
+                        pass
+
 def main():
+    """Generates CSC rst documentation from XML.
+
+    Currently goes through each topic file within each CSC folder.
+    Alphabetizing is a little weird because CSC specific topics and generic topics are located in separate
+    xml files and so each file is alphabetized seprately.
```
r-owen, line 141: seprately -> separately

I finally get it. You list specific topics first, then generic topics. Please consider stating that a bit more explicitly, something along the lines of "Component-specific topics are listed first, then generic topics. Topics are alphabetized within each group."

r-owen:
Please also edit setup.cfg so flake8 runs on this file.

Is generateCSVTable.py (not yours) still used for anything? It also has flake8 violations but I see no point to fixing it if it can be thrown out. I'll file a ticket to clean up the remaining .py files.

doc/generate_cet.py
r-owen: What does the list contain? If it is a list of specific topic names then please consider renaming it to something like specific_topics.

PR #323, opened 2020-07-09
Preview proposed solution for CAP-492

tony-johnson:
This is a "preview" of a fix for CAP-492. It will probably not be used with XML 6.1, but might be useful to start further discussion and for experimentation. The idea would be to iterate on these events for XML 6.2. These are entirely new events, so hopefully the scope for introducing problems is small.

### DIFF
--- a/sal_interfaces/CCCamera/CCCamera_Events.xml
+++ b/sal_interfaces/CCCamera/CCCamera_Events.xml
@@ -1551,4 +1551,147 @@
             <Count>240</Count>
         </item>
     </SALEvent>
+    <!-- Preview events, probably will not be used until xml 6.2 but here for comment/disscussion -->
+    <SALEvent>
+        <Subsystem>CCCamera</Subsystem>
+        <EFDB_Topic>CCCamera_logevent_imageStored</EFDB_Topic>
+        <Description>When this event is received the corresponding image should be complete in the DAQ 2-day store.
+        Once in use this will allow endReadout event to go back to indicating that the readout is complete (sequencers finished).
+        Note: all : delimited strings (aka string arrays) follow the escaping convention defined in CAP-572</Description>
+        <!-- Note, boilder plate items present on all image related events removed. TBC if this is OK -->
+        <item>
+            <EFDB_Name>imageName</EFDB_Name>
+            <Description>The imageName for this specific exposure, assigned by the camera</Description>
+            <IDL_Type>string</IDL_Type>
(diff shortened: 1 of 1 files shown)


sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -1551,4 +1551,147 @@
...
+            <Description>The file system location of each file, relative to rootFileName (: delimited).</Description>
+            <IDL_Type>string</IDL_Type>
+            <Units>second</Units>
+            <Count>9</Count>
+        </item>
+        <item>
+            <EFDB_Name>fileNode</EFDB_Name>
+            <Description>The node on which each file is written (: delimited, in same order as above).</Description>
+            <IDL_Type>string</IDL_Type>
+            <Units>second</Units>
```
mareuter, line 1619: Should be unitless.

sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -1551,4 +1551,147 @@
...
+            <Description>The common filesystem location of all files below.</Description>
+            <IDL_Type>string</IDL_Type>
+            <Units>unitless</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>relativeFileNames</EFDB_Name>
+            <Description>The file system location of each file, relative to rootFileName (: delimited).</Description>
+            <IDL_Type>string</IDL_Type>
+            <Units>second</Units>
```
mareuter, line 1612: Should be unitless.

sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -1551,4 +1551,147 @@
...
+            <Count>9</Count>
```
mareuter, line 1613: Since SAL doesn't support string arrays, I think count = 1 here.

sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -1551,4 +1551,147 @@
...
+        </item>
+        <item>
+            <EFDB_Name>fileNode</EFDB_Name>
+            <Description>The node on which each file is written (: delimited, in same order as above).</Description>
+            <IDL_Type>string</IDL_Type>
+            <Units>second</Units>
+            <Count>9</Count>
```
mareuter, line 1620: Since SAL doesn't support string arrays, I think count = 1 here.

PR #328, opened 2020-07-10
New unit test to verify string-type attributes are unitless

rbovill:


mareuter:
Is this still necessary? If it is, let's get the conflicts resolved so it can be reviewed.

tests/test_IDL_Type.py
```diff
@@ -5,11 +5,51 @@
...
 
-def check_for_issues(csc, topic):
-    jira = ""
+def check_for_issues(test, csc, topic):
+    if test == "exists" and csc == "ATPtg" and topic == "Telemetry":
+        jira = "DM-27120"
+    elif test == "exists" and csc == "MTPtg" and topic == "Telemetry":
+        jira = "DM-27121"
+    else:
+        jira = ""
```
r-owen, line 14: Please consider separate check_for_issues functions for exists and string (e.g. `check_for_type_issues` and `check_for_string_units_issues`). I think it would clarify the logic.

In fact I gently suggest moving the "strings should not have units" test to its own separate file. It's a very different test than testing that IDL_Type exists and is borderline between this file and test_Units.py.

tests/test_Units.py
```diff
@@ -59,3 +61,50 @@ def test_units(xmlfile, csc, topic):
...
+            if csc in ["ATPtg", "MTPtg"] and name.text in ts_xml.strings_with_units:
+                assert True
+            else:
+                assert unit.text in ("unitless", "dimensionless"), (
+                    csc
+                    + ": string-type attribute '"
+                    + name.text
+                    + "' has the unit '"
+                    + unit.text
+                    + "'"
```
r-owen, line 109: Please consider an f string for readability. Using !r to get the quotes (it means repr):
```
f"{csc}: string-type attribute {name.text!r} has the unit {unit.text!r}"
```

tests/test_Units.py
```diff
@@ -59,3 +61,50 @@ def test_units(xmlfile, csc, topic):
...
+    """
+    saltype = "SAL" + topic.rstrip("s")
+    # Check for known issues.
+    jira = check_for_issues(csc, topic)
+    if jira:
+        pytest.skip(
+            jira
+            + ": "
+            + str(xmlfile.name)
+            + " <Unit> fields do not conform to astropy standards."
```
r-owen, line 87: Please consider an f string for readability. The following should do it (the str is implied by the {}):
```
f"{jira}: {xmlflie.name} <Unit> fields do not conform to astropy standards."
```

r-owen (approved):
A very nice improvement.

I hope you will consider my suggestions before merging (especially trimming the list of string fields with units!).

tests/test_IDL_Type.py
```diff
@@ -5,11 +5,51 @@
...
     return jira
 
 
+@pytest.mark.parametrize("xmlfile,csc,topic", ts_xml.get_xmlfile_csc_topic())
+def test_idl_type_exists(xmlfile, csc, topic):
+    """Test that the <IDL_Type> field for topic attributes is exists at all.
```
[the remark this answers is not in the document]
rbovill: Fixed.

python/lsst/ts/xml/testutils.py
```diff
@@ -298,6 +298,31 @@
...
+"""Define string attributes that are NOT unitless"""
+strings_with_units = [
+    "azPositions",
+    "elPositions",
+    "rotPositions",
+    "mountAzString",
+    "mountDecString",
+    "mountRAString",
+    "mountRotString",
+    "mountElString",
```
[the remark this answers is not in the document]
rbovill: Updated the list.  This does not affect the check_for_issues function.

tests/test_Units.py
rbovill: Fixed

tests/test_IDL_Type.py
rbovill: I'm going to leave this on as-is.  I prefer fewer functions, and I think this is the way it's done in other test modules.  I appreciate that the 'exists' test is somewhat fluid between IDL_Type and Units, but if/when the schema tests work correctly, the exists test could be entirely removed.

tests/test_Units.py
rbovill: Fixed

PR #340, opened 2020-07-28
Remove MTArchiver from SALSubsystems.xml because of redesign

srp3rd:


rbovill:
We're going to bring this up at the next CAP meeting, on 8/4/2020.

PR #344, opened 2020-08-02
DM-25899: move doc generation code to the library

r-owen:


r-owen:
There is a test failure that appears to be from Jenkins is running too old a version of Python (older than our usual version, and older than the version used to run ts_salobj tests, which has the same code that is failing here). @rbovill do you have any suggestions for how to fix that?

tests/test_black.py
```diff
@@ -0,0 +1,41 @@
...
+        result = subprocess.run(
+            ["black", "--check", str(pkg_root), "--exclude", "version.py"],
+            capture_output=True,
+        )
+        if result.returncode != 0:
+            raise AssertionError(result.stderr)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
couger01, line 41: Is there a point to testing the black executable existence here?

r-owen: The test in question tests that all Python code is formatted with black. I added a doc string.

PR #358, opened 2020-08-21
Add undocumented field

couger01:


doc/private-topics.rst
```diff
@@ -34,6 +34,9 @@ Private Fields
         Only present for indexed SAL components.
         The SAL index of the writer.
 
+    priority
```
r-owen, line 37: Please make it clear that this is only present for event topics (or, if you prefer, "logevent" topics").

PR #361, opened 2020-08-28
Change ATArchiver and CCArchiver to Not Configurable

srp3rd:
Both ATArchiver and CCArchiver change from ConfigurableCSC to BaseCSC


mareuter (approved):
Looks good to me.

PR #369, opened 2020-10-02
Added target and motion events to CBP

couger01:
* Updated status fields to booleans instead of integers

### DIFF
--- a/sal_interfaces/CBP/CBP_Events.xml
+++ b/sal_interfaces/CBP/CBP_Events.xml
@@ -43,7 +43,7 @@
     </SALEvent>
     <SALEvent>
         <Subsystem>CBP</Subsystem>
-        <EFDB_Topic>CBP_logevent_moving</EFDB_Topic>
+        <EFDB_Topic>CBP_logevent_inPosition</EFDB_Topic>
         <Description>Reports the motion status of the encoders</Description>
         <item>
             <EFDB_Name>elevation</EFDB_Name>
--- a/sal_interfaces/CBP/CBP_Events.xml
+++ b/sal_interfaces/CBP/CBP_Events.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="utf-8"?>
+<SALEventSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://github.com/lsst-ts/ts_xml/blob/develop/schema/SALEventSet.xsd">
+    <SALEvent>
+        <Subsystem>CBP</Subsystem>
+        <EFDB_Topic>CBP_logevent_target</EFDB_Topic>
+        <Description>Reports the commanded movement.</Description>
+        <item>
+            <EFDB_Name>elevation</EFDB_Name>
+            <Description>The commanded elevation</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>degree</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>azimuth</EFDB_Name>
+            <Description>The commanded azimuth</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>degree</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>focus</EFDB_Name>
+            <Description>The commanded focus</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>millimeter</Units>
+            <Count>1</Count>
+        </item>
+        <item>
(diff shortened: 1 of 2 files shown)


sal_interfaces/CBP/CBP_Events.xml
```diff
@@ -0,0 +1,84 @@
...
+            <EFDB_Name>mask_rotation</EFDB_Name>
+            <Description>The commanded mask rotation</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>degree</Units>
+            <Count>1</Count>
+        </item>
+    </SALEvent>
+    <SALEvent>
+        <Subsystem>CBP</Subsystem>
+        <EFDB_Topic>CBP_logevent_moving</EFDB_Topic>
```
r-owen, line 46: Please consider `inPosition` instead of `moving`. Advantages:
* It tells the user what they need to know (I have arrived at the target).
* It is an event output (in some form) by most other CSCs that have actuators. Often there is one "inPosition" event that has a single field for "all actuators in position" plus and another event like this one with a separate field for each actuator. But just one even like this would, I think be fine (@tribeiro  might have a different opinion).

A separate event for "moving" is also nice to have if you actually *know* that the motors are moving. But if it's just based on "not in position" (as I _think_ is the case from reading the code in ts_CBP) then I suggest not having this event.

An event indicating the status of each actuator controller is also very useful, if you can manage that. Typically a bitmask that explains what is wrong, that includes bits for whatever you can learn, such as "forward limit switch, reverse limit switch, L3 limit switch (that cuts power to the actuator), amplifier disabled, etc. -- whatever information is actually available from the actuator controller.

If the controller already has a status word available (most of them do) then that is ideal -- that way the meaning of the bits matches what is in the actuator controller manual.

An associated enumeration with specified values is great for allowing users to know what the bits mean.
Here is an example from Rotator:
```
<Enumeration>
    ApplicationStatus_HEX_FOLLOWING_ERROR_MASK=0x00000001,
    ApplicationStatus_HEX_MOVE_COMPLETE_MASK=0x00000002,
    ApplicationStatus_COMMAND_REJECT_MASK=0x00000020,
    ApplicationStatus_SAFTEY_INTERLOCK=0x00000040,
    ApplicationStatus_SAFETY_INTERLOCK=0x00000040,
    ApplicationStatus_EXTEND_LIMIT_SWITCH=0x00000080,
    ApplicationStatus_RETRACT_LIMIT_SWITCH=0x00000100,
    ApplicationStatus_ETHERCAT_PROBLEM=0x00000200,
    ApplicationStatus_DDS_COMMAND_SOURCE=0x00000400,
    ApplicationStatus_MOTION_TIMEOUT=0x00000800,
    ApplicationStatus_DRIVE_FAULT=0x00002000,
    ApplicationStatus_SIMULINK_FAULT=0x00004000,
    ApplicationStatus_ENCODER_FAULT=0x00008000
</Enumeration>
```

sal_interfaces/CBP/CBP_Telemetry.xml
```diff
@@ -7,42 +7,42 @@
         <item>
             <EFDB_Name>panic</EFDB_Name>
             <Description>The status of the CBP.</Description>
-            <IDL_Type>float</IDL_Type>
+            <IDL_Type>boolean</IDL_Type>
```
r-owen, line 10: Please consider making `status` an event. it's all state-like information that would be more useful if reported only when it changes.

I realize this is a pre-existing issue, but I really hope you will consider changing it on this ticket branch.

r-owen: Please update the descriptions to say what the fields mean. For example, for this field I am guessing that panic= "True" means the CBP in in panic state, but the description suggests otherwise and I don't really know what a panic state is (though it sounds bad).

The others are even less clear. "azimuth=True" is good? bad? What is it telling me? The description offers no clue.

I realize this is a pre-existing issue, but I beg you to fix it on this ticket branch.

r-owen (changes requested):
Thank you for the addition. The commanded info is a great addition.

I suggest changing the motion event to inPosition, or supporting both if you have the necessary info from the actuator controllers.

Please look over the various descriptions (especially in CBP_status) and make sure they are clear and meaningful.

This may be out of scope, but:
* Most of the telemetry looks more like events, in that it has discrete state that would be easier to handle if it was output when it changed.
* Please consider a new telemetry event for the current encoder position of each actuator. That is classic telemetry and arguably the only topic that should actually be telemetry.
* Please consider removing the CBP_ prefix from all topic names. It is redundant.

r-owen (approved):
Looks great. Thank you for improving the descriptions.

PR #372, opened 2020-10-14
Update RuntimeLanguages field for Script and Test

rbovill:
* Added CPP and SALPY to the RuntimeLanguages field for the Script CSC.  SALPY is needed for the SAL python unit tests, and C++ is needed to create the SALPY library. 
* Added CPP, SALPY, Java and LabVIEW to the RuntimeLanguages list for the Test CSC. Since Test is for testing SAL, it should probably build all the libraries, and C++ and SALPY are required to run the unit tests, anyway.


r-owen (approved):
Looks fine.

PR #373, opened 2020-10-16
Feature/camera xml 6.2

tony-johnson:
Sorry but this one giant pull request for all of the changes assigned to xml 6.2, specifically:

https://jira.lsstcorp.org/issues/?jql=project%20%3D%20CAP%20AND%20fixVersion%20%3D%20%22ts_xml%206.2%22%20AND%20assignee%20in%20(tjohnson)

Let me know if you find problems.


sal_interfaces/MTCamera/MTCamera_Telemetry.xml
```diff
@@ -1,948 +1,3407 @@
...
+            <Description>Ambient Temperature</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>Celsius</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>fan1Speed</EFDB_Name>
+            <Description>Fan 1 Speed (RPM)</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units/>
```
mareuter, line 1914: Units seem to be missing here and the next two entries. Is it because RPM isn't supported by default?

sal_interfaces/MTCamera/MTCamera_Telemetry.xml
```diff
@@ -1,948 +1,3407 @@
...
+            <Description>Ambient Temperature</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>Celsius</Units>
+            <Count>1</Count>
+        </item>
+        <item>
+            <EFDB_Name>fanSpeed</EFDB_Name>
+            <Description>Fan Speed (RPM)</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units/>
```
mareuter, line 2072: Missing units here too.

sal_interfaces/MTCamera/MTCamera_Telemetry.xml
```diff
@@ -1,948 +1,3407 @@
...
+    </SALTelemetry>
+    <SALTelemetry>
+        <Subsystem>MTCamera</Subsystem>
+        <EFDB_Topic>MTCamera_focal_plane_Segment</EFDB_Topic>
+        <item>
+            <EFDB_Name>i</EFDB_Name>
+            <Description>Amplifier current</Description>
+            <IDL_Type>double</IDL_Type>
+            <Units>mA</Units>
+            <Count>3024</Count>
```
mareuter, line 3386: Corner rafts aren't included?

sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -17,7 +17,7 @@
             <Enumeration>OFFLINE_AVAILABLE,OFFLINE_PUBLISH_ONLY</Enumeration>
         </item>
         <item>
-            <EFDB_Name>transitionTimestamp</EFDB_Name>
+            <EFDB_Name>transitionTimestA</EFDB_Name>
```
mareuter, line 20: Why are we abbreviating timestamp to timestA here and below? Is there some length constraint in Java?

mareuter (commented):
- The test for Units for the ATCamera and MTCamera XML are being skipped because the have tickets against them. We need to remove the skips to see what turns up. Remove the lines associated with the CAP-318 ticket: https://github.com/lsst-ts/ts_xml/blob/develop/tests/test_Units.py#L9
- Need to do the same for the AttributeDescription test: https://github.com/lsst-ts/ts_xml/blob/develop/tests/test_AttributeDescription.py#L9
- Speaking of units, I see you mix Volt and volt for units it would be nice to use one or the other.
- Noticed that all of the attributes that had (RPM) in the description had no units. Do we need another exception for this?
- The timestamp abbreviation seems a bit odd. See inline comment.
- Is the autogeneration system going to produce output in the same order between runs? It was really hard to follow changes at times.

tony-johnson:
> * Speaking of units, I see you mix Volt and volt for units it would be nice to use one or the other.

The pull request I made on Friday accidentally had the last 3 commits missing. One of the things fixed was to consistently use Volt.

sal_interfaces/CCCamera/CCCamera_Events.xml
tony-johnson: This was also fixed by the missing 3 commits, and was a mistake introduced by replacing Amp with A (which also replaced timestamp with timestA)

sal_interfaces/MTCamera/MTCamera_Telemetry.xml
tony-johnson: I think this is correct, the corner raft boards do not have the ability to measure amplifier currents.

sal_interfaces/MTCamera/MTCamera_Telemetry.xml
tony-johnson: Fixed. I think we previously decided RPM was not a unit and should by included in the title, and that unit should be unitless (which it will be after the next push).

sal_interfaces/MTCamera/MTCamera_Telemetry.xml
tony-johnson: Fixed

tony-johnson:
> * The test for Units for the ATCamera and MTCamera XML are being skipped because the have tickets against them. We need to remove the skips to see what turns up. Remove the lines associated with the CAP-318 ticket: https://github.com/lsst-ts/ts_xml/blob/develop/tests/test_Units.py#L9

I think this is still needed for the Torr, mTorr issue. The plan was for Astropy to add support for this, I don't know if that has happened yet.

tony-johnson:
I have removed the CAP-318 exceptions, but at the price of adding VA and psia to the exception list of units not supported by astroPy

rbovill:
> > * The test for Units for the ATCamera and MTCamera XML are being skipped because the have tickets against them. We need to remove the skips to see what turns up. Remove the lines associated with the CAP-318 ticket: https://github.com/lsst-ts/ts_xml/blob/develop/tests/test_Units.py#L9
> 
> I think this is still needed for the Torr, mTorr issue. The plan was for Astropy to add support for this, I don't know if that has happened yet. We will also need to decide what to do about PSIA, which I am told is what some of the gauges measure.

Astropy has done the work and it will be included in the next release.  When that occurs, however... ¯\_(ツ)_/¯

rbovill (approved):
The changes look good to me.  I did have to make one update, to re-add the CAP-592 skip ticket in test_Units.py, as PromptProcessing has not yet been updated/redesigned.

PR #377, opened 2020-10-26
DM-26367: Remove misspelled enums.

r-owen:


rbovill (approved):
Erroneous entries removed.

PR #378, opened 2020-10-27
DM-27228: FA bump test

pkubanek:
Added bump test and kill bump test commands and bump test status event.

### DIFF
--- a/sal_interfaces/MTM1M3/MTM1M3_Commands.xml
+++ b/sal_interfaces/MTM1M3/MTM1M3_Commands.xml
@@ -805,4 +805,46 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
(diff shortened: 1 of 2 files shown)


rbovill (changes requested):
Please remove the <Alias> tags from all topic definitions.  This field was removed from the schema.

pkubanek:
> Please remove the tags from all topic definitions. This field was removed from the schema.

Can you be more specific? I am not sure what to remove. Thanks.

pkubanek:
> Please remove the tags from all topic definitions. This field was removed from the schema.

Fixed - Alias removed. As noted on slack, xmllint can do XSD validation and is used in GitHub action. It might make sense to use it in GitHub action for events etc. validations..

rbovill (approved):
Changes look good and the unit tests are passing.

PR #383, opened 2020-11-11
Add RuntimeLanguages for components based on latest build information.

tribeiro:


rbovill (changes requested):
Petr's changes need to be updated; Python should be SALPY and please remove the parentheticals.

Please re-add the languages for the Script entry; they are needed for the unit tests to run correctly.

rbovill (approved):
LGtM.

PR #384, opened 2020-11-11
Modified reported filter and grating events. Removed enumeration from…

patrickingraham:
… reportedFilterPosition and reportedDisperserPosition. Also modified the command that had text mentioning an enumeration.


tribeiro (approved):
Looks good! remember to rebase to develop before merging.

PR #389, opened 2020-11-16
Updated VERSION to 7.0.0_RC1.

rbovill:


mareuter (approved):
I assume this is to set the initial (RC) build before doing the main build. The commit message is a bit misleading since from it one would expect the version to be 7.0.0_RC1 and not 7.0.0.

rbovill:
Yes, I first do a quick RC build, then I cut the official release.  I left both in the commit history, for the sake of transparency.

PR #392, opened 2020-11-30
(DM-27642) Added description tags to the telemetry and command topics.

wvreeven:


r-owen (approved):
Looks good. I realize it's hard to be very informative, but it's nice to have the metadata.

PR #394, opened 2020-12-11
DM-27895: Overhaul the way the MT Hexapods handle compensated moves

r-owen:


teweitsai (approved):
The update looks good to me! Thanks for all the effort!

PR #396, opened 2020-12-23
DM-27996: Add minimal NewMTMount telemetry, just for the camera cable wrap and rename NewMTMount to MTMount.

r-owen:


sal_interfaces/MTMount/MTMount_Commands.xml
```diff
@@ -3,201 +3,127 @@
...
-    <Explanation>http://sal.lsst.org/SAL/Commands/MTMount_command_trackTarget.html</Explanation>
+    <Description>Specify position, velocity and time (PVT) vector for all axes. This command is immediately reported as done.</Description>
     <item>
-      <EFDB_Name>az_angle</EFDB_Name>
-      <Description>Azimuth angle</Description>
+      <EFDB_Name>azimuth</EFDB_Name>
+      <Description>Mount azimuth angle at the specified time. The allowed range is the software limits (L1), approximately -270 to 270.</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>degree</Units>
+      <Units>deg</Units>
```
wvreeven, line 56: You replace `degree` here with `deg` but on lines 32 and 39 you keep using `degree`. Perhaps replace all occurrences of `degree` with `deg`?

sal_interfaces/MTMount/MTMount_Commands.xml
```diff
@@ -3,201 +3,127 @@
...
+      <IDL_Type>long</IDL_Type>
+      <Units>dimensionless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>tracksys</EFDB_Name>
+      <Description>Defines tracking coordinate system as sidereal, non-sidereal or local.</Description>
       <IDL_Type>string</IDL_Type>
-      <Units>unitless</Units>
+      <Units>dimensionless</Units>
```
wvreeven, line 98: Why `dimensionless` instead of `unitless`? Same for line 91. `unitless` still is used in other places in this file.

sal_interfaces/MTMount/MTMount_Events.xml
```diff
@@ -1,85 +1,245 @@
...
+      <Count>1</Count>
+    </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>MTMount</Subsystem>
+    <EFDB_Topic>MTMount_logevent_axesState</EFDB_Topic>
+    <Description>State of the elevation and azimuth axes.</Description>
+    <item>
+      <EFDB_Name>elevation</EFDB_Name>
+      <Description>Elevation state, as an ActuatorState constant enum</Description>
```
wvreeven, line 67: Nitpicking but please add a trailing `.` to terminate the phrase. Same below which I assume happened due to copying and pasting.

sal_interfaces/MTMount/MTMount_Events.xml
```diff
@@ -1,85 +1,245 @@
...
+    <item>
+      <EFDB_Name>taiTime</EFDB_Name>
+      <Description>TAI date at which the commanded positions and velocities apply, as unix time.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>trackId</EFDB_Name>
+      <Description>Unique target identifier </Description>
```
wvreeven, line 191: Please replace the trailing space with a `.`

sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -74,221 +57,122 @@
...
-      <Size_in_bytes>1</Size_in_bytes>
-      <IDL_Type>string</IDL_Type>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
+    <EFDB_Topic>MTMount_azimuthDrives</EFDB_Topic>
     <item>
-      <EFDB_Name>ACW_Current_Drive_1</EFDB_Name>
-      <Description>ACW_Current_Drive_1</Description>
+      <EFDB_Name>current1</EFDB_Name>
```
wvreeven, line 62: Wouldn't it be better to replace all these currents with a single array item?

sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -297,252 +181,66 @@
...
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
+    <EFDB_Topic>MTMount_cameraCableWrap</EFDB_Topic>
     <item>
-      <EFDB_Name>Azimuth_Status_Drive_15</EFDB_Name>
-      <Description>Azimuth_Status_Drive_15</Description>
-      <Size_in_bytes>1</Size_in_bytes>
-      <IDL_Type>string</IDL_Type>
+      <EFDB_Name>driveState1</EFDB_Name>
```
wvreeven, line 186: Again, wouldn't an array be better here? Same for angle, speed and current? Perhaps it is not as important here as above where there are many more values.

sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -551,5643 +249,148 @@
...
       <Count>1</Count>
     </item>
+  </SALTelemetry>
+  <SALTelemetry>
+    <Subsystem>MTMount</Subsystem>
+    <EFDB_Topic>MTMount_elevationDrives</EFDB_Topic>
     <item>
-      <EFDB_Name>Azimuth_Surface_Temperature_Motor_12</EFDB_Name>
-      <Description>Azimuth_Surface_Temperature_Motor_12</Description>
+      <EFDB_Name>current1</EFDB_Name>
```
wvreeven, line 308: Again, wouldn't an array be better here?

wvreeven (commented):
I think readability of both the XML file and the code resulting from it would benefit greatly from using arrays instead of individual, numbered items like current, speed and angle. Also, using the same unit for similar quantities (like `deg` for angle instead of mixing `deg` and `degree`) would improve readability. Please share your thoughts on this with me before I request changes or approve the PR.

sal_interfaces/MTMount/MTMount_Commands.xml
r-owen: I'll standardize on unitless, though I prefer the former because it is Astropy's own standard.

sal_interfaces/MTMount/MTMount_Commands.xml
r-owen: Good catch. I changed the two instances of degree to deg.

sal_interfaces/MTMount/MTMount_Events.xml
r-owen: I did a global (with MTMount) correction of all Descriptions that did not end with . or ?

sal_interfaces/MTMount/MTMount_Events.xml
r-owen: Done (as part of the previous change).

sal_interfaces/MTMount/MTMount_Telemetry.xml
r-owen: Let's discuss this in parking lot. I also prefer arrays, but scalars have several important advantages:
- The EFD does not handle arrays well. Though it has plenty to deal with and we may come up with a reasonable solution some day.
- The data from the low-level controller is scalars (unfortunately, because it wastes a lot of space: each value has its own name in a json-encoded dict). It is more efficient to not have to preprocess them. When they get back from vacation I will ask them about this, but unless it is trivial for them it's probably not worth changing. A much more important change would be to publish CCW position and velocity; at present one has to know which of the two actuators is running to figure that out (all that data is in the telemetry topic, but it's messy to deal with).

wvreeven (approved):
LGTM

sal_interfaces/MTMount/MTMount_Commands.xml
```diff
@@ -3,201 +3,127 @@
...
+      <Description>Mount elevation angle velocity at the specified time.</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>degree/second</Units>
+      <Units>deg/second</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>time</EFDB_Name>
+      <EFDB_Name>taiTime</EFDB_Name>
       <Description>TAI date at which the commanded positions and velocities apply, as unix time.</Description>
```
pkubanek, line 82: as Unix time? Unix shall be written with capital U

r-owen:
I took your suggestion to use arrays for drive-specific telemetry fields. It was a great suggestion: it cleaned up the XML and the code (despite taking a bit more work to parse the telemetry from the low-level controller).

wvreeven (approved):
Thank you for introducing the arrays. The XML looks much cleaner now!

PR #398, opened 2021-01-12
DM-28293: Remove deprecated fields from the XML files

r-owen:


rbovill (approved):
Deprecated fields removed.  Non-standard units updated to meet Astropy format.

PR #399, opened 2021-01-15
MTMount: add cameraCableWrapTarget event

r-owen:


tribeiro (approved):
Looks good... A quick thought before you merge it. Consider dropping the "actual" for the actual values and just call them by a single name?

I think `position` is better than `actualPosition` and the description can clarify that it is the "actual position";

r-owen:
We talked about this and decided to stick with actualPosition because it matches usage in MTRotator and is less disruptive than changing both.

PR #400, opened 2021-01-19
DM-28318: remove deprecated fields from the schema

r-owen:


rbovill (approved):
Schema updates look good.

PR #402, opened 2021-01-25
DM-28030 Corrected a few HVAC Command IDL types and regenerated the T…

wvreeven:
…elemetry file as well.

### DIFF
--- a/sal_interfaces/HVAC/HVAC_Telemetry.xml
+++ b/sal_interfaces/HVAC/HVAC_Telemetry.xml
@@ -48,8 +48,8 @@
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>potenciaTrabajo</EFDB_Name>
-      <Description>Power Work</Description>
+      <EFDB_Name>percPotenciaTrabajo</EFDB_Name>
+      <Description>Percentage Power Work</Description>
       <IDL_Type>float</IDL_Type>
       <Units>%</Units>
       <Count>1</Count>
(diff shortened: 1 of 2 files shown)


sal_interfaces/HVAC/HVAC_Telemetry.xml
```diff
@@ -48,8 +48,8 @@
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>potenciaTrabajo</EFDB_Name>
-      <Description>Power Work</Description>
+      <EFDB_Name>percPotenciaTrabajo</EFDB_Name>
+      <Description>PERC Power Work</Description>
```
r-owen, line 52: A pre-existing condition, but could you use a few more words to explain what "Power Work" means? (I have no idea what PERC means, either, but you use it everywhere so I'm reluctant to ask you spell it out; presumably the CSC docs explain it).

r-owen (approved):
This looks fine. One small request for clarification of a doc string (which shows  up in at least 2 places) to consider.

sal_interfaces/HVAC/HVAC_Telemetry.xml
wvreeven: To be honest, I am not sure what "work" (or in Spanish "trabajo") really refers to. I asked but got an explanation that didn't explain this at all. The correct English description would be "percentage work power" but since the translations are automatically generated by my code it is hard to make a proper English translation. It is something on my TODO list but buried at the bottom of it. In any case, thanks for catching this. I made sure that PERC is replaced by "Percentage" now.

r-owen: PERC -> Percentage is a nice improvement. Thanks for doing that!

I gently suggest removing "perc" from the DDS field names in the XML, but I realize you'd need a translation dictionary (somewhere) to handle that sort of thing automatically (and there's no point to doing it manually with custom code).

wvreeven: Why remove that? It represents a percentage after all. And the translation dictionaries already are in place. As a matter of fact, that's exactly where I made the change in ts_hvac that you just reviewed.

r-owen: My argument is that it is unnecessary and makes the field name more confusing. We don't normally include units in field names (do you know of any other CSCs that do? In a quick search I could not find any).

"Percent" is just a unit and you have % in the units field, so the user already knows this is a percentage.

wvreeven: That makes sense. I'll open a new issue for this.

PR #405, opened 2021-01-26
Add hardpoints motion state enum

pkubanek:


rbovill (approved):
LGtM

sal_interfaces/MTM1M3/MTM1M3_Events.xml
```diff
@@ -3410,6 +3410,13 @@
     <item>
       <EFDB_Name>motionState</EFDB_Name>
       <Description>Hardpoint Actuator State motion State</Description>
+      <Enumeration>
+          Standby=0,
+          Chasing=1,
+          Stepping=2,
+          QuickPositioning=3,
+          FinePositiong=4
+      </Enumeration>
```
r-owen, line 3419: Please check the generated enum values in the IDL file (e.g. as generated with run make_idl_files.py MTM1M3 or I can generate it and send it to you if you prefer). I personally find the names for item-specific enums unsatisfactory so I always use global enums.

r-owen (commented):
Please consider using a global enum, instead of a topic-specific enum.

But if the constant names in the IDL look acceptable to you, then this is fine.

PR #408, opened 2021-01-28
Add PMD telemetry and Event

couger01:


sal_interfaces/PMD/PMD_Telemetry.xml
```diff
@@ -0,0 +1,15 @@
...
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsl"?>
+<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
+  <SALTelemetry>
+    <Subsystem>PMD</Subsystem>
+    <EFDB_Topic>PMD_position</EFDB_Topic>
+    <item>
+      <EFDB_Name>position</EFDB_Name>
+      <Description>The position of the device.</Description>
+      <IDL_Type>long</IDL_Type>
+      <Units>millimeter</Units>
```
r-owen, line 11: Please consider using `double` instead of `long` all positions in this XML, for several reasons:
* If this is a general purpose CSC then some devices you read may have floating-point values
* You can output `NaN` to indicate "no measurement available".

The only time this would be a bad idea is if the resolution was so fine that a double was not sufficient. That seems very unlikely for a micrometer. (In fact `float` might suffice, but I'd hate to lose information if it did not).

sal_interfaces/SALSubsystems.xml
```diff
@@ -1073,4 +1073,22 @@
...
+    <Github>https://github.com/lsst-ts/ts_pmd</Github>
+    <JenkinsTestResults>Not Available</JenkinsTestResults>
+    <RubinObsContact>Eric Coughlin</RubinObsContact>
+    <CSCDocs/>
+    <ProductOwner>Patrick Ingraham</ProductOwner>
+    <RelatedDocuments/>
+    <SoftwareLanguage>Python</SoftwareLanguage>
+    <RuntimeLanguages>IDL</RuntimeLanguages>
+    <VendorContact>Not Applicable</VendorContact>
+    <Simulator>Not Provided</Simulator>
```
r-owen, line 1091: I'm a bit surprised not to have a simulation mode. Wouldn't one be useful for exercising consumers of this data?

r-owen (changes requested):
I strongly suggest outputting the data as double instead of long. Feel free to push back if you think this unreasonable.

sal_interfaces/SALSubsystems.xml
couger01: Right. But I haven't implemented one yet along with actually writing any of the controller.

sal_interfaces/PMD/PMD_Telemetry.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsl"?>
+<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
+  <SALTelemetry>
+    <Subsystem>PMD</Subsystem>
+    <EFDB_Topic>PMD_position</EFDB_Topic>
+    <item>
+      <EFDB_Name>position</EFDB_Name>
```
r-owen, line 8: The `position` telemetry topic is a single measurement, with no identifying information. That suggest CSC will only be capable of reading a single micrometer, and that each micrometer will have its own SAL index. That is only acceptable if you are sure that each micrometer can be read completely independently of all others (e.g. one TCP/IP or USB or RS-232 interface per micrometer). In the Mitutoyo systems I used to use we had a single box that read multiple micrometers. Your design doesn't support that.

It also means constructing many domain participants, each one of which only reads a single micrometer. That is wasteful of resources (though using the shared memory model probably makes it tolerable), and keeping track of the state of many different CSCs. In other words: it doesn't scale very well.

I suspect the simplest solution in ts_xml is to add a "name" or "label" string field to this topic, and probably a timestamp (double) field as well. That way your CSC can theoretically (at least) report data for more than one micrometer. It can even report the exact time a measurement was made even if you read N micrometers at once (though if the exact time isn't crucial then it suffices to use the private_sndStamp field -- the time at which the message was sent; it just means that if you read 4 micrometers at once the times will be slightly off for 3 of them).

My big fear is that will make the data in the EFD difficult to "mine". I am assuming it is practical to filter the values by "name", but I don't know that for a fact.

r-owen (commented):
Thank you for changing the position field type.

I am worried about your plan to run one instance of the CSC per micrometer. See my comment on the position topic for details.

sal_interfaces/PMD/PMD_Telemetry.xml
couger01: You're right of course. The hub I have now reads from up to 8 devices. So should I make the count 8 for the position item or is there a more appropriate alternative?

sal_interfaces/SALSubsystems.xml
r-owen: I suggest making the entry correct for what you plan to support, as this makes for less "churn".

However, I also suggest figuring out if you plan to have a restricted set of CSCs. I am hoping you will decide to have one for MT and one for AT and that's it (unless there are non-MT/non-AT micrometers planned). If so, you could list those values in the appropriate tag here.

tribeiro: I think having "one for AT and one for MT" is probably not the correct use case here. We may have more than one running on AT and same for MT. @patrickingraham can probably comment on this.

sal_interfaces/PMD/PMD_Telemetry.xml
tribeiro: I think adding an additional "name" or "label" field would probably be more suitable. So each topic published can be associated with a "device or port".  It does make mining the EFD a bit trickier but still feasible. Both chronograph and DM's python library support filtering using a label. 

Having a vector of 8 values is probably not the best, since it restricts to the particular hardware you have available now.

r-owen: If you agree to make the CSC read multiple boxes then my suggestion is to make this a long array (e.g. on the order of 100 elements) coupled with:
* Set unused values to NaN so the EFD does not ingest them.
* Publish a list of micrometer labels as an event (not telemetry, since it is constant). These labels tell the user what each index is measuring, and which indices have no associated micrometer (blank label).
* The list of labels will have to be a single string, instead of an array of strings (SAL does not support arrays of strings).
* Use commas to separate the names and ignore whitespace after each comma. (I wish Python had direct support for splitting a string this way, but it can be done with re or by splitting on comma and then stripping each resulting name).
* If the list contains a blank label then you know that micrometer does not exist. This is important because you might not have all inputs in a given micrometer interface populated.
* Similarly I suggest making the list of labels only as long as needed, rather than requiring the configuration to list 100 names, most of which are blank.

If the PMD reads only one box then the above is still a reasonable way to go, but with a shorter list. How short? Do you assume 8 is the limit because your current interface box can only read 8 micrometers? Do you try to avoid future schema changes (e.g. in case we upgrade to a future box that can read more micrometers) by making the list longer? Probably 8 is fine, but XML schema changes are disruptive.

sal_interfaces/SALSubsystems.xml
r-owen: My point is that one CSC can easily read multiple micrometer interface boxes. It seems more natural to me to have one CSC per telescope than one per micrometer interface box. The main thing is you need to know which micrometers is which (presumably via CSC configuration).

That assumes the micrometers will tend to have fixed location, measuring a fixed thing. If you plan to move many of them around as needed that weakens my argument.

patrickingraham: the micrometer boxes use a serial connection, so basically they need a computer (RPi etc) nearby.
Also, in many cases the devices will be fixed, but certainly not all, so we should keep things flexible.

sal_interfaces/PMD/PMD_Telemetry.xml
tribeiro: Before we proceed with this discussion. 

DDS seems to support strings with unrestricted number of characters. Isn't that the case for arrays as well? It would be much better if we could dynamically adjust the size of the array instead of having an array of 1000 elements and we have to fiddle around with `nan` and all that.

couger01: Do we need to have a meeting about this and other issues?

sal_interfaces/SALSubsystems.xml
couger01: I'll change the simulator field to what I plan to support then.

sal_interfaces/PMD/PMD_Telemetry.xml
r-owen: @tribeiro SAL only supports fixed-length arrays (and does not support arrays of strings at all). I realize DDS is more flexible. Dave had some reason for imposing these restrictions. I don't know how difficult it would be to expand and I worry a bit what would the EFD entries would look like. I personally suggest living with the current SAL limitations for this CSC.

r-owen: @tribeiro @patrickingraham @couger01 we talked today and I think we agreed on the following (please give feedback for anything I got wrong):
* Use one SAL index per measurement aggregator device (e.g. Mitutoyo 8-micrometer box). That was the plan all along, we just confirmed it.
* Output measurements as an array of 8 values for now. Tiago will talk to Dave about adding support for variable-length arrays to SAL, but that will do for now.
* Output NaN for unused slots (and for readings that fail) in that 8 element array.
* Add a "metadata" event to describe the measurements. It should have the following fields (at least). Note that  I made some changes on as I wrote it up and thought about it, so definitely push back if you disagree:

  * kind: a single string describing the kind of measurement, e.g. "micrometers" or "incremental rotary encoders"
  * location: a string that gives the overall location of the measurement devices, e.g. "MTM1M3 Mirror Cell"
  * names: a string that is a of comma-separated names that describe in detail what each measurement device measures. Provide an explicit empty entry for each unused slot. Ignore whitespace after a comma. For example if location="MTM1M3 Mirror Cell" then names might be "x, y, zA, zB, zC, , , "
  * units: a string that gives the units of measurement, e.g. "um". I assume all devices output the same units so we just need one entry. If that is not a safe assumption please push back and we can make the string a comma-separated list of units. That will affect the description string for the field, but not the name or type of the field.

sal_interfaces/PMD/PMD_Events.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
+<SALEventSet xmlns:xsi="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsd">
+  <SALEvent>
+    <Subsystem>PMD</Subsystem>
+    <EFDB_Topic>PMD_logevent_metadata</EFDB_Topic>
+    <item>
+      <EFDB_Name>kind</EFDB_Name>
```
patrickingraham, line 8: I'd prefer "type" but I imagine you avoided that word on purpose?
maybe "instrument" or "instrument_function" ?
I imagine you don't want to use just "function"?

sal_interfaces/PMD/PMD_Events.xml
```diff
@@ -0,0 +1,36 @@
...
+    <item>
+      <EFDB_Name>location</EFDB_Name>
+      <Description>Overall location of the device.</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>names</EFDB_Name>
+      <Description>Comma-separated list of names that details what each device measures.</Description>
```
patrickingraham, line 23: I don't think I understand this....

sal_interfaces/PMD/PMD_Events.xml
```diff
@@ -0,0 +1,36 @@
...
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>units</EFDB_Name>
+      <Description>The unit of the devices.</Description>
```
patrickingraham, line 30: you mean the units of the values published by the device?
Does this have to do with the comma separated list above? I suppose we're assuming that each device cannot publish two columns of information in different units?

r-owen: It is the units of the reported measurements. One value for all measuring devices (for a given instance/SAL index of PMD). For example we will presumably set our Mitutoyo micrometers to report measurements in um, in which case this string would be "um".

This assumes we don't have some weird hybrid system where some devices attached to a box measure in um and some in degrees or.... I sure hope we can assume that. For now all we have is Mitutoyo micrometers.

r-owen (approved):
Thank you for the changes. This looks good to me.

r-owen:
I do have one more request that I hope you will consider: please add `hubType` to `PMD_logevent_metadata`. It's an important part of the CSC. In the CSC it's an enum, and that would make sense here, though it means any time you add a new hub type it requires adding an enumeration to the XML. I think we can live with that because adding hubs will be quite rare.

I wonder if we need both "kind" and "hubType"? At the moment I think they tell us the same thing. Perhaps `hubType` would do for now, as an integer and an associated enum (with only one value at the moment).

sal_interfaces/PMD/PMD_Events.xml
r-owen: In the CSC code you have hub_type which seems to serve the same purpose. As I said in a general comment I just added perhaps hubType as an int here and add an enumeration (with just one value at the moment) would do the job. I suspect you could get rid of "kind".

couger01: I renamed kind to hubType and added the enumeration as suggested by @r-owen.

r-owen:
I withdraw the suggestion about hubType. Let's just make `kind` be the same information as a string. We can add an enum later, or replace "kind", should you end up with multiple kinds of measurement hubs and decide that an enum would be convenient.

r-owen:
I suggest you merge this now without waiting for the CSC work.

PR #410, opened 2021-02-02
DM-28420: MTRotator: add command "fault"

r-owen:


teweitsai (approved):
This new xml command looks very good and useful. Thanks for the update!

PR #411, opened 2021-02-04
Update MTAOS commands based on tstn-026.

tribeiro:


sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
     <item>
-      <EFDB_Name>directoryPath</EFDB_Name>
-      <Description>Absolute directory path to calibration products.</Description>
-      <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
-      <Units>unitless</Units>
+      <EFDB_Name>ra</EFDB_Name>
+      <Description>Right Ascension of the field in ICRS coordinates.</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>hourangle</Units>
```
teweitsai, line 22: Why the `ra` use the `hourangle` and `decl` use the `degree`? Why not use the same unit?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
+      <IDL_Type>float</IDL_Type>
       <Units>degree</Units>
       <Count>1</Count>
     </item>
     <item>
       <EFDB_Name>filter</EFDB_Name>
-      <Description>Active filter type. Use 1-6 for u, g, r, i, z, and y. Use 7 for reference.</Description>
-      <IDL_Type>int</IDL_Type>
+      <Description>Name of the filter used in the observation.</Description>
+      <IDL_Type>string</IDL_Type>
```
teweitsai, line 42: I think if the type is `string`, it is better to write down the available options: u, g, r, i, z, y, and ref.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
-      <Units>degree</Units>
-      <Count>1</Count>
-    </item>
-    <item>
-      <EFDB_Name>userGain</EFDB_Name>
-      <Description>Gain value between 0 and 1. Set to -1 to ignore user gain.</Description>
-      <IDL_Type>double</IDL_Type>
+      <EFDB_Name>mode</EFDB_Name>
+      <Description>Enumeration specifying the wfs mode.</Description>
+      <IDL_Type>long</IDL_Type>
```
teweitsai, line 49: If the type is `long` or `Enumeration`, it is better to write down the details of enumeration.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTAOS</Subsystem>
-    <EFDB_Topic>MTAOS_command_processIntraExtraWavefrontError</EFDB_Topic>
-    <Description>Process the wavefront data collection by camera piston.</Description>
+    <EFDB_Topic>MTAOS_command_preProcess</EFDB_Topic>
+    <Description>Pre-process image for WEP.</Description>
```
teweitsai, line 57: I think we agreed to write down the details instead of just using the vague words of "Pre-process image".

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
     <item>
-      <EFDB_Name>intraVisit</EFDB_Name>
-      <Description>Intra-focal image visit number.</Description>
-      <IDL_Type>int</IDL_Type>
+      <EFDB_Name>visitId</EFDB_Name>
+      <Description>Image visit id number. This is a unique identifier for each image and consists of the date of the observation plus an integer iterator, e.g. 2020103000040 for the 40th image taken in October, 30, 2020.</Description>
```
teweitsai, line 60: What is the max of image number? 40 or 400 or more than this?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
       <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>MTAOS</Subsystem>
+    <EFDB_Topic>MTAOS_command_runWEP</EFDB_Topic>
+    <Description>Process wavefront data.</Description>
```
teweitsai, line 76: I think we agreed to give more description of this. For example, if the user does not issue the `selectSources()` before, but only this command, the MTAOS will look for the (`ra`, `decl`, `pa`, `filter`) in the header of image files.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
     <item>
-      <EFDB_Name>extraDirectoryPath</EFDB_Name>
-      <Description>Absolute directory path to extra-focal wavefront data collection.</Description>
-      <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
+      <EFDB_Name>visitId</EFDB_Name>
+      <Description>Image visit id number. This is a unique identifier for each image and consists of the date of the observation plus an integer iterator, e.g. 2020103000040 for the 40th image taken in October, 30, 2020.</Description>
```
teweitsai, line 79: What is the max of image number? 40 or 400 or more than this?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
-      <Description>Active filter type. Use 1-6 for u, g, r, i, z, and y. Use 7 for reference.</Description>
-      <IDL_Type>int</IDL_Type>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
-    <item>
-      <EFDB_Name>cameraRotation</EFDB_Name>
-      <Description>Camera rotation angle.</Description>
+      <EFDB_Name>wf</EFDB_Name>
+      <Description>Array with wavefront errors. Indexes are considered to be ANSI (https://en.wikipedia.org/wiki/Zernike_polynomials#OSA/ANSI_standard_indices).</Description>
```
teweitsai, line 112: Do we need this argument when the `runWEP()` is applied already? The `runOFC()` can use the result from `runWEP()`. It is better to have some clear explanation here. When should we use this and when we do not need it.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
-      <Description>Camera rotation angle.</Description>
-      <IDL_Type>double</IDL_Type>
-      <Units>degree</Units>
-      <Count>1</Count>
-    </item>
     <item>
       <EFDB_Name>userGain</EFDB_Name>
       <Description>Gain value between 0 and 1. Set to -1 to ignore user gain.</Description>
-      <IDL_Type>double</IDL_Type>
+      <IDL_Type>float</IDL_Type>
```
teweitsai, line 94: It is the `runOFC()` needs the `userGain`, not the `runWEP()`.

teweitsai (approved):
The update looks good to me. I left some comments. Please consider them before the merge. Thanks!

sal_interfaces/MTAOS/MTAOS_Commands.xml
bxin: good catch.

sal_interfaces/MTAOS/MTAOS_Commands.xml
bxin: right. we definitely need more than 1x19 numbers...

bxin (changes requested):
also suggest -
change
evt_wavefrontError.annularZernikePoly
to
evt_wavefrontError.annularZernikeCoeff

sal_interfaces/MTAOS/MTAOS_Commands.xml
teweitsai: Why we could use the same unit of `hourangle` or `degree` for both of them?

tribeiro: that's how astronomers use RA and declination...

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: Although I agree with the idea of using enums, there are a couple of drawback to doing that. 

Once is that we would be tied to the enums to add support to filters. If eventually we have different filters installed in the camera (something that may happen in the future) we would have to update the enums as well. 

The other "issue" is that, if this is an interface enum, the source of truth must be the xml and not `ts_wep`... that means, we need to add the definition to the xml file and that will have to be added to ts_idl as well.

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: I think we will have to wait to write down the details for when we know what they are. At this point it is a bit early to add details here.

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: there's no max...

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: I think this level of detail is better done in the code rather than in the xml interface. In any case, I think we can add more details here once we know what we are actually implementing.

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: there is no limit.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,121 @@
...
-      <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
-    <item>
-      <EFDB_Name>fieldRA</EFDB_Name>
-      <Description>Right ascension of telescope boresight.</Description>
-      <IDL_Type>double</IDL_Type>
+      <EFDB_Name>decl</EFDB_Name>
```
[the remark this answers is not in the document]
tribeiro: `dec` is a reserved word in DDS.. so we can't use it.

sal_interfaces/MTAOS/MTAOS_Commands.xml
teweitsai: I do not think so. Instead, for a problem, I think if we could ask the question clearly, we had solved the half of it already. I still insist a good design before the implementation is critical.

tribeiro: Not sure I understand your point. We know we will have to do "pre-processing" to those images. What are all these processes? I don't know. I know it will involve ISR and probably some other tasks. We will only know when we are actually processing real data.

I don't think a good design is one that specify every detail beforehand, but one that is flexible enough to evolve with our understanding of the problem.

teweitsai: Sure. Maybe this is the best option at this moment. Thanks!

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -9,7 +9,7 @@
   <SALCommand>
     <Subsystem>MTAOS</Subsystem>
     <EFDB_Topic>MTAOS_command_issueCorrection</EFDB_Topic>
-    <Description>Issue the wavefront correction to subsystems based on the most recently measured wavefront error.</Description>
+    <Description>Issue the wavefront correction to subsystems based on the most recently ofc state.</Description>
```
bxin, line 12: suggest: "the most recent OFC outputs".

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -54,14 +54,21 @@
   <SALCommand>
     <Subsystem>MTAOS</Subsystem>
     <EFDB_Topic>MTAOS_command_preProcess</EFDB_Topic>
-    <Description>Pre-process image for WEP.</Description>
+    <Description>Pre-process image for WEP, e.g., ISR.</Description>
```
bxin, line 57: suggest: "Pre-process image for WEP. The outputs of this command are donut images that are ready for curvature wavefront sensing".

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -73,7 +80,7 @@
   <SALCommand>
     <Subsystem>MTAOS</Subsystem>
     <EFDB_Topic>MTAOS_command_runWEP</EFDB_Topic>
-    <Description>Process wavefront data.</Description>
+    <Description>Process wavefront data, either full focal plane mode (Comcam and LSSTCam) or wavefront errors. If only visitId is provided it assumes wavefront error mode. If visitId and extraId is provided it assumes full focal plane mode. MTAOS will determine if it is ComCam or LSSTCam based on the input images. If useOCPS=True, processing is done remotely using the OCPS CSC. For LSSTCam full array mode only OCPS processing is supported. Before running this command it is advisable to run selectSources and preProcess commands on the input image(s), otherwise the MTAOS will have to run those processes before running wep.</Description>
```
bxin, line 83: "wavefront errors" -> "wavefront sensors"
"wavefront error mode"->"wavefront sensor mode".

bxin (approved):
a few more minor comments on wording. 
Good to go! Thanks.

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,159 @@
...
+      <Description>Image visit id number. This is a unique identifier for each image and consists of the date of the observation plus an integer iterator, e.g. 2020103000040 for the 40th image taken in October, 30, 2020.</Description>
+      <IDL_Type>long</IDL_Type>
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>extraVisit</EFDB_Name>
-      <Description>Extra-focal image visit number.</Description>
-      <IDL_Type>int</IDL_Type>
+      <EFDB_Name>useOCPS</EFDB_Name>
```
teweitsai, line 71: What is "OPCS"? Full name in description?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,159 @@
...
+  <SALCommand>
+    <Subsystem>MTAOS</Subsystem>
+    <EFDB_Topic>MTAOS_command_runWEP</EFDB_Topic>
+    <Description>Process wavefront data, either full focal plane mode (Comcam and LSSTCam) or wavefront sensors. If only visitId is provided it assumes wavefront sensor mode. If visitId and extraId is provided it assumes full focal plane mode. MTAOS will determine if it is ComCam or LSSTCam based on the input images. If useOCPS=True, processing is done remotely using the OCPS CSC. For LSSTCam full array mode only OCPS processing is supported. Before running this command it is advisable to run selectSources and preProcess commands on the input image(s), otherwise MTAOS will have to run those processes before running wep.</Description>
     <item>
-      <EFDB_Name>extraDirectoryPath</EFDB_Name>
-      <Description>Absolute directory path to extra-focal wavefront data collection.</Description>
-      <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
+      <EFDB_Name>visitId</EFDB_Name>
```
teweitsai, line 90: If the `extraId` is given, the `visitId` will be the intra-focal or extra-focal image? Can you add this in the description?

teweitsai: You put in the `extraId` already. Sorry!

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,159 @@
...
+      <IDL_Type>long</IDL_Type>
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>cameraRotation</EFDB_Name>
-      <Description>Camera rotation angle.</Description>
-      <IDL_Type>double</IDL_Type>
-      <Units>degree</Units>
+      <EFDB_Name>useOCPS</EFDB_Name>
```
teweitsai, line 104: What will happen if the user put `useOCPS=true` in `MTAOS_command_preProcess()` and put the value of `false` in this command (`MTAOS_command_runWEP()`)? And vice versa?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,159 @@
...
-      <Units>unitless</Units>
-      <Count>10</Count>
+      <Units>micron</Units>
+      <Count>19</Count>
     </item>
     <item>
-      <EFDB_Name>filter</EFDB_Name>
-      <Description>Active filter type. Use 1-6 for u, g, r, i, z, and y. Use 7 for reference.</Description>
-      <IDL_Type>int</IDL_Type>
+      <EFDB_Name>userGain</EFDB_Name>
```
teweitsai, line 149: What should be the order of this command and `MTAOS_command_runOFC()`? Are they independent? If not, what will happen if the users put two different `userGain` values?

sal_interfaces/MTAOS/MTAOS_Events.xml
```diff
@@ -402,40 +315,26 @@
...
-    <EFDB_Topic>MTAOS_logevent_wepWarning</EFDB_Topic>
-    <Description>Warning of wavefront estimation pipeline.</Description>
+    <EFDB_Topic>MTAOS_logevent_wepDuration</EFDB_Topic>
+    <Description>Duration of time to process the wavefront image data.</Description>
     <item>
-      <EFDB_Name>timestamp</EFDB_Name>
-      <Description>Time at which an event is recorded.</Description>
-      <IDL_Type>double</IDL_Type>
+      <EFDB_Name>calcTime</EFDB_Name>
+      <Description>Average of calculation time.</Description>
```
teweitsai, line 322: If this is a event instead of telemetry right now, we should publish the average calculation time or just the latest calculation time? Which one is more useful? I felt if this is a event now, I will prefer to use the latest calculation time.

sal_interfaces/MTAOS/MTAOS_Events.xml
```diff
@@ -402,40 +315,26 @@
...
-    <EFDB_Topic>MTAOS_logevent_ofcWarning</EFDB_Topic>
-    <Description>Warning of optical feedback control.</Description>
+    <EFDB_Topic>MTAOS_logevent_ofcDuration</EFDB_Topic>
+    <Description>Duration of time to calculate the degree of freedom.</Description>
     <item>
-      <EFDB_Name>timestamp</EFDB_Name>
-      <Description>Time at which an event is recorded.</Description>
-      <IDL_Type>double</IDL_Type>
+      <EFDB_Name>calcTime</EFDB_Name>
+      <Description>Average of calculation time.</Description>
```
teweitsai, line 334: If this is a event instead of telemetry right now, we should publish the average calculation time or just the latest calculation time? Which one is more useful? I felt if this is a event now, I will prefer to use the latest calculation time.

teweitsai (approved):
I have some questions for the use of `userGain` and calculation time may need to discuss. This update is very useful and simplify the xml a lot! Great job!

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: For the LSSTCam full array mode, `preProcess` and `runWEP` will only be supported for `useOCPS=True`. So if you send either of them with `useOCPS=False`, the command is rejected. Then, if you send `runWEP` without a previous `preProcess`, the pre-processing will be done before WEP. We should probably send a warning message in this case. 

For ComCam full array I think the command should run either way. In this case, doing `preProcess` with `useOCPS=False` and `runWEP` with `useOCPS=True` will be the same as if just executing `runWEP` with `useOCPS=True`, and vice-versa. In either case, we should probably send a warning message and it could be a special message saying that preProcess was run with a different `useOCPS` setup. 

I am weary to *forcibly require* `preProcess` to run with the same `useOCPS` because there are cases where we may simply not do `preProcess` and just want to execute `runWEP` and let it deal with all the steps. 

For LSST corner wavefront sensor there's no `preProcess` step so, that's it.

tribeiro: I added some information on the xml about this.

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: I think this made it here by (cut and paste) mistake... I'll remove it.

sal_interfaces/MTAOS/MTAOS_Commands.xml
teweitsai: I do not see where is the added information. BTW, it is hard to trace the update if you continue to use the force-push to replace the latest commit. At each time, the reviewer will need to re-read the commit to look for the updated part. Thanks!

tribeiro: It's in the description, one line 105

```
<Description>Use OCS-Controlled Pipeline System (OCPS) to process data remotely? Ideally this should match the value of a previous preProcess command. If runWEP and preProcess run with different useOCPS setups, runWEP will run normally but will take longer as it will have to repeat the steps of preProcess. A warning message will inform the users of these conditions.</Description>
```

teweitsai: Got it. Thanks!

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: removed

sal_interfaces/MTAOS/MTAOS_Events.xml
tribeiro: Updated to "Latest execution time"

sal_interfaces/MTAOS/MTAOS_Events.xml
tribeiro: Updated to "Latest execution time"

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,152 @@
...
-    <item>
-      <EFDB_Name>directoryPath</EFDB_Name>
-      <Description>Absolute directory path to calibration products.</Description>
-      <IDL_Type>string</IDL_Type>
-      <IDL_Size>1</IDL_Size>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
+    <EFDB_Topic>MTAOS_command_rejectCorrection</EFDB_Topic>
+    <Description>Reject the most recent wavefront correction.</Description>
```
sandrinethomas, line 17: What happens when you reject a correction?

sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -3,244 +3,152 @@
...
   </SALCommand>
   <SALCommand>
     <Subsystem>MTAOS</Subsystem>
-    <EFDB_Topic>MTAOS_command_processWavefrontError</EFDB_Topic>
```
sandrinethomas, line 43: Is that the ISR only?

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: It will reverse the OFC state to the one before the calculation was done.

sal_interfaces/MTAOS/MTAOS_Commands.xml
tribeiro: Uhm... this comment seems to have been made in a command that was removed. Are you asking about the `processWavefrontError` or the `selectSources` command?

sal_interfaces/MTAOS/MTAOS_Events.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
 <SALEventSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsd">
+  <Enumeration>FilterType_u,FilterType_g,FilterType_r,FilterType_i,FilterType_z,FilterType_y</Enumeration>
```
teweitsai, line 4: This is inconsistent with the IDL now:
https://github.com/lsst-ts/ts_idl/pull/57/commits/1b6c0edd9deabd9cd9cb1091d464bf6bdff177fa

tribeiro: updated in both places.

sandrinethomas (approved):
I only had a couple of comments.
Thank you!

sal_interfaces/MTAOS/MTAOS_Commands.xml
bxin: perhaps add what you say here to the xml description as well? (since it wasn't obvious to me either)

bxin:
Since you haven't merged this yet - it is kind of hard to tell the difference between resetCorrection and rejectCorrection (at least to me). resetCorrection can easily feel like it refers to the (relative) correction calculated for this visit.
Would renaming resetCorrection to zeroCorrection be better?

PR #415, opened 2021-02-15
Add instructions on how to add enumerations to SAL XML

couger01:


doc/sal_constraints_and_recommendations.rst
```diff
@@ -452,6 +453,14 @@ e.g. for ATTCS in C++:
...
     ATTCS_someEvent_SpecificSetA (datum specific)
 
+Define it using the following format.
+
+1. In ``CSCName_Events.xml`` create an ``Enumeration`` tag just before the first ``SAlEvent`` tag in the file.
+2. Fill out the Enumeration with comma delimited values using the following naming format ``itemName_value``.
+3. Repeat for as many enumerations as you require.
+
+.. warning:: Some enumerations are defined inside of the topic themselves. This is a legacy method and no longer recommended or necessary.
+
```
pkubanek, line 463: Shouldn't you mention you shall (optionally) add enumerations values as constant to ts_idl/python/lsst/ts/idl/enums/<CSCName.py? At least before that would be auto-generated?

doc/sal_constraints_and_recommendations.rst
```diff
@@ -452,6 +453,15 @@ e.g. for ATTCS in C++:
     ATTCS_shared_SimpleSetA (declared globally)
     ATTCS_someEvent_SpecificSetA (datum specific)
 
+Define it using the following format.
+
+1. In ``CSCName_Events.xml`` create an ``Enumeration`` tag just before the first ``SAlEvent`` tag in the file.
+2. Fill out the Enumeration with comma delimited values using the following naming format ``itemName_value``.
+3. Repeat for as many enumerations as you require.
+4. Optionally, add the enumerations to ``ts_idl`` by creating a ``lsst.ts.idl.enums.{csc_name}`` module.
```
tribeiro, line 461: Can you remove the "Optionally"?

couger01: Done.

pkubanek: Well they might be cases (C++ only) when this is not needed. And I believe we target to autogenerate those (which shall not be so difficult..it's just XML parsing, I can add this feature if requested to make_idl_files.py)

couger01: But it is helpful for ts_salobj remotes.

pkubanek (approved):
Looks good. Thanks for adding ts_idle enums to description.

PR #416, opened 2021-02-15
Add Guider interface.

tribeiro:


sal_interfaces/Guider/Guider_Commands.xml
```diff
@@ -0,0 +1,61 @@
...
+<SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
+  <SALCommand>
+    <Subsystem>Guider</Subsystem>
+    <EFDB_Topic>Guider_command_startGuiding</EFDB_Topic>
+    <Description>Start guiding using the input information. The command returns as soon as guiding starts, before the first interation is completed. If it fails to find a guide star in the provided guider region, the CSC will go to fault state.</Description>
+    <item>
+      <EFDB_Name>roiX</EFDB_Name>
+      <Description>Vector with 8 elements specifying the center of the guider ROI window in the x-axis (in pixels) for each guider sensor. If an element of the array is zero, that guider will be ignored.</Description>
+      <IDL_Type>long</IDL_Type>
+      <Units>unitless</Units>
```
pkubanek, line 12: Please change unit to pix. That's legal unit for Astropy:

```
In [19]: u.Unit('pixel')                                                                                 
Out[19]: Unit("pix")
```

Not only for this, but for all other unitless pixels. You might then omit "(in pixels)".

sal_interfaces/Guider/Guider_Commands.xml
```diff
@@ -0,0 +1,61 @@
...
+      <Count>8</Count>
+    </item>
+    <item>
+      <EFDB_Name>roiY</EFDB_Name>
+      <Description>Vector with 8 elements specifying the center of the guider ROI window in the x-axis (in pixels) for each guider sensor. If an element of the array is zero, that guider will be ignored.</Description>
```
pkubanek, line 17: Please replace x-axis with y-axis (in Description).

sal_interfaces/Guider/Guider_Commands.xml
```diff
@@ -0,0 +1,61 @@
...
+    <item>
+      <EFDB_Name>roiHeight</EFDB_Name>
+      <Description>Vector with 8 elements specifying the height (window size in the y-axis) of the guider ROI for each guider sensor (in pixels). For each valid entry in roiX/roiY (roiX,roiY &gt; 0) height must respect the ROI constrains (see requirement CA-TS-GDR-ICD-0006 in LSE-66).</Description>
+      <IDL_Type>long</IDL_Type>
+      <Units>unitless</Units>
+      <Count>8</Count>
+    </item>
+    <item>
+      <EFDB_Name>expTime</EFDB_Name>
+      <Description>Exposure time in milliseconds. All guider sensors will use the same exposure time (CA-TS-GDR-ICD-0034 LSE-66). Limits must respect detector integration time constrains (see requirement CA-TS-GDR-ICD-0034 in LSE-66).</Description>
```
pkubanek, line 38: The Astropy preffered abbervation is ms:
```
In [21]: u.Unit('millisecond')                                                                           
Out[21]: Unit("ms")
```
also consider if ms are rellay needed. Will s not suffice?

tribeiro: Nice, I'll update to `ms`... I'm torn about using `seconds`... although I agree with you it would suffice, the ICD specifies exposure time in milliseconds.

sal_interfaces/Guider/Guider_Commands.xml
tribeiro: done

sal_interfaces/Guider/Guider_Commands.xml
tribeiro: nice catch! done.

sal_interfaces/Guider/Guider_Commands.xml
```diff
@@ -0,0 +1,61 @@
...
+      <IDL_Type>float</IDL_Type>
+      <Units>milliseconds</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>binning</EFDB_Name>
+      <Description>Binnig factor. All guider sensors will use the same binning factor. Limits must respect detector bining constrains (see requirement CA-TS-GDR-ICD-0042 in LSE-66).</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>milliseconds</Units>
```
pkubanek: Should be unitless or pix. It's hard to see how that can be ms. Is that both vertical & horizontal? Aren't there separate values for vertical and horizontal innings? If that's both axis, unit shall probably be int/long (binning usually doesn't work with pixel rational parts).

sal_interfaces/Guider/Guider_Commands.xml
pkubanek: Also center can be on fractional pixels. Assume you choose odd size for ROI (e.g. 3x3), your pixel center would be 1.5 in x and y (placed on origin). It's probably better to specify ROI using roiXleft,roiXright,roiYbottom,roiYupper coordinates.

sal_interfaces/Guider/Guider_Events.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
+<SALEventSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsd">
+  <Enumeration>GuiderStatus_Stopped,GuiderStatus_Starting,GuiderStatus_Guiding,GuiderStatus_Error</Enumeration>
```
pkubanek, line 4: That might be too narrow. Guider CCD needs to readout the image. Using GuiderStatus_Exposing,GuiderStatus_Readout might give more data about guiding progress. Not sure if that's needed, but might allow showing progress in GUI/EUI.

sal_interfaces/Guider/Guider_Events.xml
```diff
@@ -0,0 +1,60 @@
...
+  <SALEvent>
+    <Subsystem>Guider</Subsystem>
+    <EFDB_Topic>Guider_logevent_guidingStatus</EFDB_Topic>
+    <Description>Event published when a guiding operatation starts, stops or resumes. It contains the parameters used to configure the guider (see startGuiding command) and the current status.</Description>
+    <item>
+      <EFDB_Name>roiX</EFDB_Name>
+      <Description>Vector with 8 elements specifying the center of the guider ROI window in the x-axis (in pixels) for each guider sensor. If an element of the array is zero, that guider will be ignored.</Description>
+      <IDL_Type>long</IDL_Type>
+      <Units>unitless</Units>
+      <Count>8</Count>
```
pkubanek, line 14: See above. Unit shall be pix, either change type to float or specify ROI as a rectangle (topleft/upperright pixels)

sal_interfaces/Guider/Guider_Events.xml
```diff
@@ -0,0 +1,60 @@
...
+      <Description>Vector with 8 elements specifying the height (window size in the y-axis) of the guider ROI for each guider sensor (in pixels). For each valid entry in roiX/roiY (roiX,roiY &gt; 0) height must respect the ROI constrains (see requirement CA-TS-GDR-ICD-0006 in LSE-66).</Description>
+      <IDL_Type>long</IDL_Type>
+      <Units>unitless</Units>
+      <Count>8</Count>
+    </item>
+    <item>
+      <EFDB_Name>expTime</EFDB_Name>
+      <Description>Exposure time in milliseconds. All guider sensors will use the same exposure time (CA-TS-GDR-ICD-0034 LSE-66). Limits must respect detector integration time constrains (see requirement CA-TS-GDR-ICD-0034 in LSE-66).</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>milliseconds</Units>
```
pkubanek: ms?

sal_interfaces/Guider/Guider_Events.xml
```diff
@@ -0,0 +1,60 @@
...
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>binning</EFDB_Name>
+      <Description>Binnig factor. All guider sensors will use the same binning factor. Limits must respect detector bining constrains (see requirement CA-TS-GDR-ICD-0042 in LSE-66).</Description>
+      <IDL_Type>float</IDL_Type>
```
pkubanek, line 47: not millisends, shall be int/long

sal_interfaces/Guider/Guider_Events.xml
pkubanek: Also what happens if you lost guider star(s)? Shall that not be reflected in some flags? Will then status switch to "waiting for guide * to re-appear"?

sal_interfaces/Guider/Guider_Telemetry.xml
```diff
@@ -0,0 +1,23 @@
...
+      <IDL_Type>double</IDL_Type>
+      <Units>arcsec</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>y</EFDB_Name>
+      <Description>Guider offset in the (instrument) y-axis direction.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>arcsec</Units>
+      <Count>1</Count>
```
pkubanek, line 20: Please either add topic for Guider_mmOffsets specifying the offset in mm in focal plane. From mm offsets from center as measured on detector (which is independent of binning) and given guider plane rotation (and telescope F), you can convert the offset to arcsec useful for guiding. Those arcsec offsets better be specified in telescope coordinates, e.g. alt and el. You can also record mm offsets per guider (so count=8), as then you run some magic to calculate the single offset you send to the TCS.

sal_interfaces/Guider/Guider_Commands.xml
```diff
@@ -0,0 +1,61 @@
...
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsl"?>
+<SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
+  <SALCommand>
+    <Subsystem>Guider</Subsystem>
+    <EFDB_Topic>Guider_command_startGuiding</EFDB_Topic>
+    <Description>Start guiding using the input information. The command returns as soon as guiding starts, before the first interation is completed. If it fails to find a guide star in the provided guider region, the CSC will go to fault state.</Description>
+    <item>
+      <EFDB_Name>roiX</EFDB_Name>
+      <Description>Vector with 8 elements specifying the center of the guider ROI window in the x-axis for each guider sensor. If an element of the array is zero, that guider will be ignored.</Description>
+      <IDL_Type>long</IDL_Type>
```
pkubanek, line 11: Center can be on fractional pixels. Assume you choose odd size for ROI (e.g. 3x3), your pixel center would be 1.5 in x and y (placed on origin). It's probably better to specify ROI using roiXleft,roiXright,roiYbottom,roiYupper coordinates.

pkubanek (changes requested):
Please see comment on problem with odd size ROI and center pixel.

sal_interfaces/Guider/Guider_Events.xml
tribeiro: The guider operates at a really high rate... I think if we where to capture the "status" with such fine details the state would be switching back and forth so fast it would be difficult to follow up. 

For the case where it loose the guide star, I think the best option would be to go to fault and then the status would be Error. If, internally, the Guider has the capability of waiting for a couple seconds before recovering the guide start, the status would remain "Guiding" and warning messages would log the condition. 

At the same time, if we ever feel we need a finer detail, we can update the enum set.

sal_interfaces/Guider/Guider_Telemetry.xml
tribeiro: This is the single offset sent to the TCS... So, any magic in combining those 8 measures need to be done here. But that's a good point about using mm in focal plane.

sal_interfaces/Guider/Guider_Commands.xml
tribeiro: good point!

tribeiro:
I think I implemented all your comments, please, take a look again..

pkubanek (approved):
Thanks for changes.

PR #420, opened 2021-02-17
DM-28848: Escape || bars around MTHexapod events/commands description.

couger01:


sal_interfaces/MTHexapod/MTHexapod_Commands.xml
```diff
@@ -24,7 +24,7 @@
     <Description>Set position and orientation limits.</Description>
     <item>
       <EFDB_Name>maxXY</EFDB_Name>
-      <Description>Maximum |X| and |Y|</Description>
+      <Description>Maximum X and Y</Description>
```
r-owen, line 27: These bars indicate the absolute value, so I would prefer to keep them if possible. Why are they triggering warnings in the documentation generation step? Can you fix that step, instead?

r-owen (commented):
Please find a different solution, if practical.

r-owen (approved):
Please change the `|` to `\|`. Then feel free to merge.

If I'd known that Description strings were interpreted as reStructuredText I'd have been happy to do this myself. I thought it was a bug in the documentation generator.

PR #421, opened 2021-03-01
DM-28934: The WeatherStation CSC units have been fixed.

wvreeven:


sal_interfaces/WeatherStation/WeatherStation_Telemetry.xml
```diff
@@ -8,21 +8,21 @@
...
-      <Units>unitless</Units>
+      <Units>%</Units>
       <Count>1</Count>
     </item>
     <item>
       <EFDB_Name>pressure</EFDB_Name>
       <Description>The pressure outside.</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>unitless</Units>
+      <Units>pa</Units>
```
pkubanek, line 25: Meteorologists are using hPa as the standard unit for atmospheric pressure. Maybe switch to hPa?

wvreeven: Correct. I will update that because the WeatherStation indeed outputs hPa.

sal_interfaces/WeatherStation/WeatherStation_Telemetry.xml
```diff
@@ -339,21 +339,21 @@
       <EFDB_Name>paAvg1M</EFDB_Name>
       <Description>1 minute average value</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>pa</Units>
+      <Units>hPa</Units>
```
pkubanek, line 342: See for pressure value..this is hPa (preferred?), above is pa. pa isn't astropy unit, correct is Pa.

pkubanek:
I would probably as well switch various n minutes/hours averages into events. Assuming those are computed by CsC at some interval. If low level HW produces those values per readout, leave them in telemetry.

wvreeven:
The low level hardware produces the averages so we will keep them in the telemetry.

rbovill (approved):
Units are updated from unitless.  New yaml script added to enforce proper formatting of Python scripts.  All looks good.

PR #422, opened 2021-03-01
DM-28981: Update ApplicationStatus enums for MTHexapod and MTRotator

r-owen:


wvreeven (approved):
Looks good to me. Thank you for removing those trailing zeros. This is much clearer.

PR #425, opened 2021-03-08
DM-28997: A few minor changes to telemetry and command names.

wvreeven:


r-owen (approved):
This looks acceptable but: I am starting to worry about how much churn there is and how many almost identical commands exist. I suggest you consider having just *one* enable command that specifies the system to enable or disable, and *one* configure command that specifies the name of the device and the configuration in json format.

I am also a bit concerned that the control system only has "day" and "night" settings. What controls when we switch between these? Is two enough? I can imagine possibly wanting day, 1 hour before opening, and night.

r-owen:
Two further thoughts:
* If you do consolidate the enable commands into one: make the "what to enable" a bit mask. Then you can enable many devices at once. In which case you should have separate "enable" and "disable" commands, or possibly two separate bit mask arguments (enable and disable) for the one command.
* For the configure commands you could retain type safety, if you like, by having one command per device type (again with a bit mask). It'll take a few lines of code to detect attempts to configure the wrong kind of device (an inappropriate bit mask).

wvreeven:
I very much like the idea of consolidating the enable and configure commands and created a ticket for it: DM-29135.

PR #430, opened 2021-03-30
M2 and rotator VMS

pkubanek:


dmills-lsst (approved):
ok

tribeiro (changes requested):
Can you add a "description" field to each topic and to each topic attribute explaining what they are?

I would also like to suggest you change the topics to be "sensor specific" instead of "component specific". So instead of `MTVMS_m1m3` and `MTCMS_m2` you could have `MTVMS_sensor3C` and  `MTVMS_sensor6C`, respectively.

sal_interfaces/MTVMS/MTVMS_Telemetry.xml
```diff
@@ -14,161 +14,276 @@
     <item>
       <EFDB_Name>sensor1XAcceleration</EFDB_Name>
```
tribeiro, line 15: DM has a nice general naming convention (https://developer.lsst.io/cpp/style.html#general-naming-conventions) we should try to adopt as much as possible.

One of the guidelines state:

> The fundamental quantity being described should appear first in the name, with modifiers concatenated afterward. 

I would suggest changing from `sensor1Xacceleration` to `accelerationXSensor1` (preferred) or `xAccelerationSensor1`. Same throughout the other topics attributes.

pkubanek:
> Can you add a "description" field to each topic and to each topic attribute explaining what they are?

There is description already, for all fields I can found. Can you please be specific for which topics/items description is missing?

> I would also like to suggest you change the topics to be "sensor specific" instead of "component specific". So instead of `MTVMS_m1m3` and `MTCMS_m2` you could have `MTVMS_sensor3C` and `MTVMS_sensor6C`, respectively.

If I do that and add enum indicating source, would it be possible in EFD make independent columns, so one doesn't have to filter with enum, but can choose only collumns to display data?

sal_interfaces/MTVMS/MTVMS_Telemetry.xml
pkubanek: done, renamed to accelerationXSensor1 etc.

tribeiro:
> > Can you add a "description" field to each topic and to each topic attribute explaining what they are?
> 
> There is description already, for all fields I can found. Can you please be specific for which topics/items description is missing?
>

I see the description for the topics attributes but not for the topic themselves. 

e.g.: https://github.com/lsst-ts/ts_xml/blob/77cb59855edadd66ec8116289ff6be3b13d2632d/sal_interfaces/MTVMS/MTVMS_Telemetry.xml#L5-L6

I would expect something like:

```
    <Subsystem>MTVMS</Subsystem>
    <EFDB_Topic>MTVMS_m1m3</EFDB_Topic>
    <Description>M1M3 (3-channel) accelerometer measurements.</Description>
```

> 
> > I would also like to suggest you change the topics to be "sensor specific" instead of "component specific". So instead of `MTVMS_m1m3` and `MTCMS_m2` you could have `MTVMS_sensor3C` and `MTVMS_sensor6C`, respectively.
> 
> If I do that and add enum indicating source, would it be possible in EFD make independent columns, so one doesn't have to filter with enum, but can choose only collumns to display data?

Currently, I don't think that is possible but I agree that it is a reasonable request. I would have to talk to Russell and folks from square about that, but I do like that idea.

Thanks for updating the names of the topics attributes.

tribeiro (approved):
This looks considerably cleaner! thanks for the updates! looks great!

PR #431, opened 2021-03-30
DM-29451: Update the black version in .pre-commit-config.yaml

r-owen:


rbovill (approved):
Black version updated to v20.  Looks good.

PR #432, opened 2021-03-31
Torr is now an acceptable Astropy unit. Removed the workaround from t…

rbovill:
…ests/test_Units.py


r-owen (approved):
Looks good. A nice improvement.

PR #433, opened 2021-04-05
DM-29570: Remove unused "value" fields from generic topics

r-owen:


rbovill (approved):
Unused attributes were removed.

PR #437, opened 2021-04-07
DM-29135: Consolidated the enable and configuration commands.

wvreeven:


sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,15 +1,20 @@
...
-  <Enumeration>DetailedState_DisabledState,DetailedState_EnabledState,DetailedState_FaultState,DetailedState_OfflineState,DetailedState_StandbyState</Enumeration>
+  <Enumeration>DisabledState_DisabledState,DisabledState_EnabledState,DisabledState_FaultState,DisabledState_OfflineState,DisabledState_StandbyState</Enumeration>
+  <Enumeration>ChillerId_chiller01=1,ChillerId_chiller02=2,ChillerId_chiller03=3</Enumeration>
+  <Enumeration>CrackId_crack01=1,CrackId_crack02=2</Enumeration>
+  <Enumeration>FancoilId_fancoil01=1,FancoilId_fancoil02=2,FancoilId_fancoil03=3,FancoilId_fancoil04=4,FancoilId_fancoil05=5,FancoilId_fancoil06=6,FancoilId_fancoil07=7,FancoilId_fancoil08=8,FancoilId_fancoil09=9,FancoilId_fancoil10=10,FancoilId_fancoil11=11,FancoilId_fancoil12=12</Enumeration>
+  <Enumeration>ManejadoraId_manejadoraSblanca=1,ManejadoraId_manejadoraSlimpia=2</Enumeration>
+  <Enumeration>ManejadoraLowerId_manejadoraLower01=1,ManejadoraLowerId_manejadoraLower02=2,ManejadoraLowerId_manejadoraLower03=3,ManejadoraLowerId_manejadoraLower04=4</Enumeration>
```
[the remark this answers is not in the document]
r-owen: Please consider having a unique ID value for each device. The big advantage is type safety: it is impossible to use the wrong bitmask and get surprised by which device is configured -- instead it'll reference an invalid device. (C++ and Python can enforce type safety without doing this, but SAL transmits these values as an integer).

I see two basic approaches to this:
* Use the index of the bit in the enable/disable mask as the ID. Thus the ID values and enable/disable mask have an obvious 1:1 relation and you could even compute one from the other in ts_idl. With my example just above you'd have chiller01=1, chiller02=2, chiller03=3, fancoil01=4, ... fancoil12=15, ...
* Use the enable/disable bit values "as is". This gives you exactly one enum class and exactly one way to identify a device. This is clearly the best approach if you think you ever might want to support one configuration command configuring multiple devices. Otherwise...I have mixed feelings.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,15 +1,20 @@
...
   <SALEvent>
     <Subsystem>HVAC</Subsystem>
     <EFDB_Topic>HVAC_logevent_detailedState</EFDB_Topic>
+    <Description>Event for the detailedState.</Description>
```
r-owen, line 14: Do you have detailed state for this CSC? I don't see the enums above. Please consider deleting this event.

I admit I am biased. I have never written a CSC that outputs detailed state because I have never found a single integer to be sufficient for usefully expressing details of internal state.

r-owen (changes requested):
This looks like a really useful improvement.

My main concern is that one has to enable/disable _all_ devices with the new enable command, and I think users may find that quite limiting compared to being able to change select devices and leave the others alone.

Also please add an enum for the bits in the enable/disable mask(s).

Also (optional) please consider unique IDs for the configuration commands.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,23 +1,92 @@
...
+  <Enumeration>
+    ManejadoraId_mSblanca=1,
+    ManejadoraId_mSlimpia=2
+  </Enumeration>
+  <Enumeration>
+    ManejadoraLowerId_m01=1,
+    ManejadoraLowerId_m02=2,
+    ManejadoraLowerId_m03=3,
+    ManejadoraLowerId_m04=4
+  </Enumeration>
```
r-owen, line 91: I *strongly* suggest you just use the HvacTopicId enum for configuration commands.

This gives one standard way to refer to a device. It also addresses the issue I raised last time: that your configuration commands aren't safe because a user can use the wrong enum.

A very nice way to handle this would be to assign a range of values for each category of device (Tekniker does this). For example:
```
    DeviceId_Chiller01P01=101,
    DeviceId_Chiller02=102,
    DeviceId_chiller03=103,
    DeviceId_Crack01P02=201,
    DeviceId_Crack02P02=202,
    DeviceId_Fancoil01P02=301,
    ...
    DeviceId_Fancoil12P02=312,
    DeviceId_ManejadoraLower01P05=401
    ...  
```

Then if you need to send indices 1-N to the lower level device, you can subtract the appropriate multiple of 100 -- information you can easily store in another enum, or a Python dict. Of course multiples of 100 aren't necessary at all, but something like that does allow adding and subtracting devices without having to change other values, and it is quite readable.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,23 +1,92 @@
...
+    HvacTopicId_vea12P05=0X001E,
+    HvacTopicId_vea13P05=0X001F,
+    HvacTopicId_vea14P05=0X0020,
+    HvacTopicId_vea15P05=0X0021,
+    HvacTopicId_vea16P05=0X0022,
+    HvacTopicId_vea17P05=0X0023,
+    HvacTopicId_vec01P01=0X0024,
+    HvacTopicId_vin01P01=0X0025,
+    HvacTopicId_vex03LowerP04=0X0026,
+    HvacTopicId_vex04CargaP04=0X0027
```
r-owen, line 57: Please consider removing the Hvac prefix because it does not add much, if any, information.

Perhaps DeviceId instead? "Topic" seems an odd name for something that you enable and configure. But if that's the terminology the vendor uses, "TopicId" would do.

sal_interfaces/HVAC/HVAC_Events.xml
r-owen: I gently suggest you use base 10 for this enum and reserve hex values for bit masks. Using hex for bit masks is is a common convention, probably because that make it obvious when a large number is a power of 2. For other sequences, base 10 is fine and arguably more readable.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,23 +1,92 @@
...
-      <Count>1</Count>
-    </item>
-  </SALEvent>
+  <Enumeration>
+    SummaryState_DisabledState,
+    SummaryState_EnabledState,
+    SummaryState_FaultState,
+    SummaryState_OfflineState,
+    SummaryState_StandbyState
+  </Enumeration>
```
r-owen, line 10: Please delete the SummaryState enum. It duplicates an enumeration that ts_sal already provides.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,23 +1,92 @@
...
+  <Enumeration>
+    DisabledState_DisabledState,
+    DisabledState_EnabledState,
+    DisabledState_FaultState,
+    DisabledState_OfflineState,
+    DisabledState_StandbyState
+  </Enumeration>
```
r-owen, line 17: Please delete this. I suspect it may have started out as DetailedState, but surely you are not using it.

r-owen (changes requested):
Please delete the SummaryState and DisabledState enums. Sorry I missed those earlier.

I strongly suggest a single set of enums for device IDs that you use for both enable/disable and configuration. That makes the configuration commands safer because there is no risk of using the wrong enum class.

sal_interfaces/HVAC/HVAC_Events.xml
wvreeven: I agree and will make the changes. "DeviceId" is fine. I can make the translation in the CSC.

sal_interfaces/HVAC/HVAC_Events.xml
wvreeven: Done

sal_interfaces/HVAC/HVAC_Events.xml
wvreeven: Done

sal_interfaces/HVAC/HVAC_Events.xml
wvreeven: A good suggestion and thanks for the short explanation on Slack. Done.

sal_interfaces/HVAC/HVAC_Commands.xml
```diff
@@ -3,68 +3,39 @@
 <SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
   <SALCommand>
     <Subsystem>HVAC</Subsystem>
-    <EFDB_Topic>HVAC_command_chiller01P01_enable</EFDB_Topic>
-    <Description>Enable command for the chiller01P01 device.</Description>
+    <EFDB_Topic>HVAC_command_enable_device</EFDB_Topic>
+    <Description>Enable command for a single HVAC device.</Description>
```
r-owen, line 7: Please consider "Enable a single HVAC device" (active voice).

And same for the disable command.

sal_interfaces/HVAC/HVAC_Commands.xml
```diff
@@ -3,68 +3,39 @@
...
     <item>
-      <EFDB_Name>comandoEncendido</EFDB_Name>
-      <Description>Command Started</Description>
-      <IDL_Type>boolean</IDL_Type>
+      <EFDB_Name>device_id</EFDB_Name>
+      <Description>The ID indicating which single device needs to be enabled. The IDs can be found in the DeviceId enumeration.</Description>
```
r-owen, line 10: I think "single" is redundant here. I realize you originally coded a bitmask, but that version was never released, and DeviceId is not a bitmask, so I don't think anyone will be confused.

sal_interfaces/HVAC/HVAC_Commands.xml
```diff
@@ -75,56 +46,15 @@
...
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
-  </SALCommand>
-  <SALCommand>
-    <Subsystem>HVAC</Subsystem>
-    <EFDB_Topic>HVAC_command_crack01P02_enable</EFDB_Topic>
-    <Description>Enable command for the crack01P02 device.</Description>
+    <EFDB_Topic>HVAC_command_config_cracks</EFDB_Topic>
+    <Description>Configuration command for command for a single HVAC device in the CRACK device group.</Description>
```
r-owen, line 50: Please consider "Configure a single Crack device". (Or CRACK if it's an acronym). If you can provide a brief translation/description of what a Crack device is, that would also be helpful.

and similarly for the other configure commands.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,18 +1,45 @@
...
+    DeviceId_vea13P05=0X025F,
+    DeviceId_vea14P05=0X0260,
+    DeviceId_vea15P05=0X0261,
+    DeviceId_vea16P05=0X0262,
+    DeviceId_vea17P05=0X0263,
+    DeviceId_vea01P01=0X02BD,
+    DeviceId_vec01P01=0X02BE,
+    DeviceId_vin01P01=0X02BF,
+    DeviceId_vex03LowerP04=0X0321,
+    DeviceId_vex04CargaP04=0X0322
```
r-owen, line 43: I gently suggest using decimal instead of hex so users know that this is not a bitmask.

r-owen (approved):
This looks very nice and is a very nice simplification of the API.

I had a few minor suggestions I hope you will consider.

sal_interfaces/HVAC/HVAC_Commands.xml
wvreeven: Yes I keep forgetting about that. Done.

sal_interfaces/HVAC/HVAC_Commands.xml
wvreeven: Good suggestion. I removed it here and also for the enable and disable commands.

sal_interfaces/HVAC/HVAC_Commands.xml
wvreeven: Even better, thanks! Done.

sal_interfaces/HVAC/HVAC_Events.xml
wvreeven: Of course. Sorry, this slipped by my attention and was a left over from a previous iteration. Done.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,18 +1,243 @@
...
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>HVAC</Subsystem>
+    <EFDB_Topic>HVAC_logevent_chiller_configured</EFDB_Topic>
+    <Description>Event indicating that the CHILLER devices were configured.</Description>
+    <item>
+      <EFDB_Name>device_mask</EFDB_Name>
```
r-owen, line 62: I strongly suggest this be device_id, not a bit mask (and similarly for the other configuration events). Surely it is not common for multiple devices to have the same configuration?

Similarly for the other configuration events.

Also this indicates the current configuration I hope you will report it startup (in addition to when it is changed). So I suggest a name such as "chillerConfiguration" and a description such as "Configuration for a chiller".

In any case please use camelCase for the names (after HVAC_logevent_) since that is our standard.

r-owen (changes requested):

sal_interfaces/HVAC/HVAC_Events.xml
r-owen: I was probably too hard on your decision to use a bitmask and I apologize. It can certainly work. I worry it'll be significantly more work to write such messages because (in order to use the format efficiently) you'll want to accumulate sets of devices that have the same configuration. But Python sets or possibly dicts could be used for that purpose. I also worry it'll make reading the data from the EFD a lot harder.

wvreeven: I don't think you were too hard. I modified it to a single device ID.

sal_interfaces/HVAC/HVAC_Commands.xml
```diff
@@ -156,53 +86,15 @@
   </SALCommand>
   <SALCommand>
     <Subsystem>HVAC</Subsystem>
-    <EFDB_Topic>HVAC_command_crack02P02_enable</EFDB_Topic>
-    <Description>Enable command for the crack02P02 device.</Description>
+    <EFDB_Topic>HVAC_command_config_fancoils</EFDB_Topic>
```
r-owen, line 89: Please use configFancoils (or fancoilsConfig) for the topic name. Or at least check with Tiago if you want to use snake_case instead.

r-owen (changes requested):

sal_interfaces/HVAC/HVAC_Commands.xml
wvreeven: I chose configFancoils and also modified the enable and disable commands to enableDevice and disableDevice.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,18 +1,243 @@
...
+    DeviceId_vec01P01=702,
+    DeviceId_vin01P01=703,
+    DeviceId_vex03LowerP04=801,
+    DeviceId_vex04CargaP04=802
+  </Enumeration>
   <SALEvent>
     <Subsystem>HVAC</Subsystem>
-    <EFDB_Topic>HVAC_logevent_detailedState</EFDB_Topic>
+    <EFDB_Topic>HVAC_logevent_deviceEnabled</EFDB_Topic>
+    <Description>Event indicating that a device was enabled.</Description>
```
r-owen, line 48: "a device" implies this is for a single device (like the command).

Please consider "Report which devices are enabled" or similar.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,18 +1,243 @@
...
+      <Description>Setpoint Active</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>deg_C</Units>
+      <Count>1</Count>
+    </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>HVAC</Subsystem>
+    <EFDB_Topic>HVAC_logevent_crackConfiguration</EFDB_Topic>
+    <Description>Event indicating that the CRACK devices were configured.</Description>
```
r-owen, line 79: "were configured" sounds like yes/no and "devices were" suggests it applies to multiple devices.
Please consider "Configuration of a CRACK device" or similar.

Similarly for the other configuration events.

sal_interfaces/HVAC/HVAC_Events.xml
```diff
@@ -1,18 +1,243 @@
...
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>HVAC</Subsystem>
+    <EFDB_Topic>HVAC_logevent_chillerConfiguration</EFDB_Topic>
+    <Description>Event indicating that the CHILLER devices were configured.</Description>
+    <item>
+      <EFDB_Name>device_id</EFDB_Name>
+      <Description>The ID indicating which device was configured. The mapping of the ID to the corresponding device can be found in the DeviceId enumeration with IDs in the 100 range.</Description>
```
r-owen, line 63: I find this quite confusing as I can't tell if you mean the user to provide a DeviceId enum or somehow remap that to some other value.

Please consider "Device ID; one of the DeviceId_chiller enums"

and similarly for the others.

r-owen (approved):
Thanks for fixing the issues.

I have some requests to clarify descriptions that I hope you will consider, but I'll leave it to you whether to make those changes.

PR #446, opened 2021-04-27
DM-28586: Removed trailing word 'perc' from HVAC command, event and t…

wvreeven:
…elemetry items with a 'percentage' unit.


r-owen (approved):
Looks great. Thank you for making this change.

PR #455, opened 2021-06-10
DM-30695: Minor tweaks to ATMonochromator and ATWhiteLightSource

r-owen:


wvreeven (approved):
In general the changes look good. One nit-picking concern: some of the descriptions start with a lower case and some with an upper case letter. At some time it would be nice to have them all changed to upper case.

r-owen:
That is a good point. And some end with "." and some do not. In general I think this XML needs a significant cleanup; for instance I see no need to have "timestamp" values for events (though they can be useful for telemetry) nor do I see any point to publishing the system time (errors in the system clock can be caught by comparing private_sndStamp to private_rcvStamp).

PR #457, opened 2021-06-11
Add new item to endOfImageTelemetry: emulatedImage

tony-johnson:
When using emulated DAQ, add emulatedImage name to the end of image telemetry, to make it possible for @menanteau to add it to header service if desired.


mareuter (approved):
Looks fine to me.

PR #458, opened 2021-06-11
DM-30711: Added new commands and modified telemetry of MTDome.

wvreeven:


sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -80,42 +86,42 @@
...
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_closeShutter</EFDB_Topic>
-    <Description>Close the shutter.</Description>
+    <Description>Close the shutter. This should also be used for an emergency stop.</Description>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_openShutter</EFDB_Topic>
-    <Description>Open the shutter.</Description>
+    <Description>Open the shutter. This should also be used for an emergency stop.</Description>
```
r-owen, line 99: I suggest reverting this change. This operation is not a "stop" and I think it would be a very unusual emergency that required _opening_ the shutter.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -80,42 +86,42 @@
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stopLouvers</EFDB_Topic>
-    <Description>Stop all louvers.</Description>
+    <Description>Stop all louvers. This should also be used for an emergency stop.</Description>
```
r-owen, line 89: I find this additional text confusing -- both on this command and all the other device-specific stop commands below. What sort of emergency requires stopping only the louvers (or other single device)?

My first thought would be to simply remove the text. If you expand the description of the "stop" and "goStationary" commands then that should suffice.

Two other options include changing the description to something like "Quickly stop all louvers" or to state that this is one of the things the "stop" command does. Both imply "there is no quicker way to stop the louvers", which I suspect is the point.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -59,6 +59,12 @@
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_park</EFDB_Topic>
+    <Description>Move all components to park position and engage the brakes and locking pins.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_command_goStationary</EFDB_Topic>
+    <Description>Stop all motion and engage the brakes. For an emergecy stop, use the stop command.</Description>
```
r-owen, line 67: Please explain *why* this command is not suitable for an emergency. My *guess* is that it stops azimuth motion more slowly than the `stop` command, and that all other actuators stop equally quickly for both commands. If that is the case then all those extra "This should also be used for an emergency stop" descriptions you added below are probably unnecessary.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -80,42 +86,42 @@
...
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stopLouvers</EFDB_Topic>
-    <Description>Stop all louvers.</Description>
+    <Description>Stop all louvers. This should also be used for an emergency stop.</Description>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_closeShutter</EFDB_Topic>
-    <Description>Close the shutter.</Description>
+    <Description>Close the shutter. This should also be used for an emergency stop.</Description>
```
r-owen, line 94: I suggest reverting this change. Closing is not any kind of "stop".

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -80,42 +86,42 @@
...
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stopEl</EFDB_Topic>
-    <Description>Stop elevation (light/wind screen) motion.</Description>
+    <Description>Stop elevation (light/wind screen) motion. This should also be used for an emergency stop.</Description>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stop</EFDB_Topic>
-    <Description>Stop all motion.</Description>
+    <Description>Stop all motion. This should also be used for an emergency stop.</Description>
```
r-owen, line 119: Please rework this text to include a reference to goStationary and an explanation of why one might use one or the other of these two commands. Again, my guess is that this one stops the azimuth in a way that is abrupt and hard on the dome. If so, please consider renaming "goStationary" to "stop" and this command to "estop" or "emergencyStop". I realize it's potentially confusing to have the command differ from the low-level controller, but this is the _public_ API and it should be as clear as possible (and "goStationary" is just weird).

r-owen (changes requested):
A few suggestions that are on the edge of significant enough for "Request changes". My main concern is the "goStationary" command -- I am hoping you can rename that to "stop" and come up with a better name for the emergency stop command -- such as emergencyStop. (In my opinion matching EIE is much less important than providing a clear user interface, though it'd be even better to get both if EIE is willing to change).

I would also like to see clarification of the descriptions for the `stop` and `goStationary` command, as well, so a user understands how they differ.

r-owen:
After talking to @wvreeven here's the difference:
* goStationary halts the dome but leaves the brakes off
* stop halts the actuators and applies the brakes (apparently in that order, rather than simultaneously, so it stops just as gently as goStationary).

So I suggest combining the two commands into one: `stop` with a boolean parameter `brakes`. I think that would be much clearer to users.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: Yeah, that was a bit too enthusiastic copying and pasting. Reverted.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: Change reverted.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: As discussed on Slack, I removed this command and added an `engage_brakes` parameter to all stop commands.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -95,27 +103,55 @@
...
+      <IDL_Type>boolean</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stop</EFDB_Topic>
-    <Description>Stop all motion.</Description>
+    <Description>Stop all motion. Issue this command also to disengage the brakes and locking pins when in PARKED state.</Description>
```
r-owen, line 142: I find this is a confusing command because it does three loosely related things:
* Stop the dome, optionally applying the brakes, once stopped.
* Engage or disengage the brakes (if already stopped).
* Disengage the locking pins (if parked).

It's probably too late to change, but I think the API would be much clearer if disengaging the locking pins was a separate command.

In any case, I think the description needs to be clearer, to fully explain these three operations. Here are two ideas:

Stop any existing motion and then engage or disengage the brakes. This command also disengages locking pins, so the recommended way to unpark the mode is to issue this command with engageBrakes=false.

or possibly:

For all subsystems that are moving: stop the motion and then optionally apply the brakes. For all subsystems that are not moving: disengage the locking pins (unparking the dome) and engage or disengage the brakes.

r-owen (approved):
This looks great. I did suggest a few changes that I hope you will consider before merging:
* engage_brakes needs to be engageBrakes to meet our standard. It's a great field name -- very clear.
* Expanding the description for the stop command would be much appreciated. That command does a lot.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -80,7 +81,14 @@
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stopLouvers</EFDB_Topic>
-    <Description>Stop all louvers.</Description>
+    <Description>Stop all louvers. Issue this command also to disengage the brakes and locking pins when in PARKED state.</Description>
```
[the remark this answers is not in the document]
wvreeven: There is a `MTDome_command_stop` which can be used for the whole system in one go. I modified the description once more.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -95,27 +103,55 @@
...
+      <Count>1</Count>
+    </item>
   </SALCommand>
   <SALCommand>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_command_stop</EFDB_Topic>
-    <Description>Stop all motion.</Description>
+    <Description>Stop all motion. Issue this command also to disengage the brakes and locking pins when in PARKED state.</Description>
+    <item>
+      <EFDB_Name>engage_brakes</EFDB_Name>
```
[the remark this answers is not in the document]
wvreeven: Yes I forgot about that, thanks. Done.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: I adopted the second suggestion.

PR #459, opened 2021-06-14
DM-30741: Add a job id event to the OCPS.

ktlim:


sal_interfaces/OCPS/OCPS_Events.xml
```diff
@@ -28,4 +28,31 @@
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>OCPS</Subsystem>
+    <EFDB_Topic>OCPS_logevent_job_id</EFDB_Topic>
```
mareuter, line 33: I have a preference for this being called: OCPS_logevent_jobId.

sal_interfaces/OCPS/OCPS_Events.xml
```diff
@@ -28,4 +28,31 @@
...
+    <item>
+      <EFDB_Name>event_name</EFDB_Name>
```
mareuter, line 35: I think in order to maintain consistency in the attributes either write them as:
- eventName
- seqNum
- jobId

or
- event_name
- seq_num
- job_id

mareuter (approved):
A couple of naming suggestions, but otherwise this is fine.

sal_interfaces/OCPS/OCPS_Events.xml
ktlim: The name `job_id` was for parallelism with the existing `job_result`.

sal_interfaces/OCPS/OCPS_Events.xml
ktlim: Similarly, the field in `job_result` is `job_id`.  `seqNum` was for parallelism with its underlying source: `private_seqNum`, as that is an important detail needed for its usage.  But if we don't mind usage like `event.seq_num == trigger.private_seqNum`, I'm happy to change it.

mareuter: OK. If there is precedent for they way things are names, then they can all stay the way you originally wrote them.

PR #462, opened 2021-06-15
DM-30762: Replace Units=dimensionless with unitless

r-owen:


rbovill:
Since we're removing **dimensionless** has a valid Unit choice, I think we should update the test_Units.py unit test to remove the workaround for **dimensionless**.

mareuter (approved):
Looks good. Thanks for the changes.

PR #468, opened 2021-06-30
DM-29906: Implemented all commands described in the ICD in the DREAM …

wvreeven:
…command XML.

### DIFF
--- a/sal_interfaces/DREAM/DREAM_Commands.xml
+++ b/sal_interfaces/DREAM/DREAM_Commands.xml
@@ -3,11 +3,31 @@
 <SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
   <SALCommand>
     <Subsystem>DREAM</Subsystem>
-    <EFDB_Topic>DREAM_command_setEnabled</EFDB_Topic>
-    <Description>Enable or disable DREAM.</Description>
+    <EFDB_Topic>DREAM_command_resume</EFDB_Topic>
+    <Description>Indicate that DREAM is permitted to resume automated operations.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_openHatch</EFDB_Topic>
+    <Description>Open the hatch if DREAM has evaluated that it is safe to do so.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_closeHatch</EFDB_Topic>
+    <Description>Close the hatch.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_stop</EFDB_Topic>
+    <Description>Immediately stop operations and close the hatch.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_readyForData</EFDB_Topic>
+    <Description>Inform DREAM that Rubin Observatory is ready to receive data or not.</Description>
     <item>
-      <EFDB_Name>enabled</EFDB_Name>
-      <Description>DREAM enabled or disabled.</Description>
+      <EFDB_Name>ready</EFDB_Name>
+      <Description>Ready to receive data or not.</Description>
       <IDL_Type>boolean</IDL_Type>
       <Units>unitless</Units>
       <Count>1</Count>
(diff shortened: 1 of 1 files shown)


sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_closeHatch</EFDB_Topic>
+    <Description>Close the hatch.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>DREAM</Subsystem>
+    <EFDB_Topic>DREAM_command_stop</EFDB_Topic>
+    <Description>Immediately stop operations and close the hatch.</Description>
```
teweitsai, line 22: Will this be different from `SALGeneric_command_abort`?

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
     <Subsystem>DREAM</Subsystem>
-    <EFDB_Topic>DREAM_command_operate</EFDB_Topic>
-    <Description>Operate DREAM, meaning taking images, or not.</Description>
+    <EFDB_Topic>DREAM_command_readyForData</EFDB_Topic>
+    <Description>Inform DREAM that Rubin Observatory is ready to receive data or not.</Description>
     <item>
-      <EFDB_Name>operating</EFDB_Name>
-      <Description>DREAM operating, meaning taking images, or not.</Description>
+      <EFDB_Name>ready</EFDB_Name>
+      <Description>Ready to receive data or not.</Description>
```
teweitsai, line 30: If the Rubin Observatory is not ready yet, why does it want to issue the  `DREAM_command_readyForData` with the `ready=False`?

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
-    <Description>Request a data product from DREAM.</Description>
     <item>
-      <EFDB_Name>product_name</EFDB_Name>
-      <Description>The name of the DREAM data product.</Description>
-      <!-- DM-29091 TODO: Replace with an enum as soon as the data products have been defined.  -->
-      <IDL_Type>string</IDL_Type>
-      <Units>unitless</Units>
+      <EFDB_Name>humidity</EFDB_Name>
+      <Description>The current humidity.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 55: Do we need the humidity value to be `float`?

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <Units>%</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>timestamp</EFDB_Name>
-      <Description>The TAI timestamp of observation producing the DREAM data product.</Description>
-      <IDL_Type>string</IDL_Type>
+      <EFDB_Name>wind_speed</EFDB_Name>
+      <Description>The current wind speed.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 62: Do we need this value to be `float`? It looks like the page 9 of ICD only says the type to be `number`.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <Units>m/s</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>wind_direction</EFDB_Name>
+      <Description>The current wind direction.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 69: Do we need this value to be `float`? It looks like the page 9 of ICD only says the type to be `number`.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
-      <IDL_Type>string</IDL_Type>
+      <EFDB_Name>wind_speed</EFDB_Name>
+      <Description>The current wind speed.</Description>
+      <IDL_Type>int</IDL_Type>
+      <Units>m/s</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>wind_direction</EFDB_Name>
+      <Description>The current wind direction.</Description>
```
teweitsai, line 68: It might be good to say this is azimuth (based on page 9 of ICD), although I do not think this is clear enough to define the direction.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <IDL_Type>int</IDL_Type>
+      <Units>deg</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>pressure</EFDB_Name>
+      <Description>The current air pressure.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 76: Do we need this value to be `float`? It looks like the page 9 of ICD only says the type to be `number`.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <Units>Pa</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>rain</EFDB_Name>
+      <Description>The current rain.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 83: Do we need this value to be `float`? It looks like the page 9 of ICD only says the type to be `number`.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <Units>mm</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>cloudcover</EFDB_Name>
+      <Description>The current cloudcover.</Description>
+      <IDL_Type>int</IDL_Type>
```
teweitsai, line 90: Do we need this value to be `float`? It looks like the page 9 of ICD only says the type to be `number`.

sal_interfaces/DREAM/DREAM_Commands.xml
```diff
@@ -3,61 +3,98 @@
...
+      <Units>%</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>safe_observing_conditions</EFDB_Name>
+      <Description>Are the conditions safe for observing or not.</Description>
+      <IDL_Type>boolean</IDL_Type>
```
teweitsai, line 97: If the condition is good enough to do the observation, why do we need to send the values such as the rain, humidity, etc.?

teweitsai (approved):
This update looks good to me. I only have some logic problems for the curiosity. Thanks!

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: Yes,the command has an interesting name. In general `ready=True` will be sent. If there is some internal problem (network to the rest of the world down for instance) then `ready=False` will be sent. In that case DREAM will stop sending data (and store it internally) but keep observing.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: The data are coming from the WeatherStation and I checked that XML and indeed it is float. I will modify that.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: The data are coming from the WeatherStation and I checked that XML and indeed it is int. Perhaps we should update the ICD.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: The data are coming from the WeatherStation and I checked that XML and indeed it is int. Perhaps we should update the ICD.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: I added that to the description. Why wouldn't that be clear enough?

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: The data are coming from the WeatherStation and I checked that XML and indeed it is float. I will modify that.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: The data are coming from the WeatherStation and I checked that XML and indeed it is float. I will modify that.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: There is no cloud cover info in WeatherStation. Still, I will modify this to float.

sal_interfaces/DREAM/DREAM_Commands.xml
teweitsai: I think the (x, y, z) is needed to define the direction. This is 3D but the azimuth is only for one direction. But maybe I am wrong in this field. Thanks!

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: Rain probably will not be used and it may disappear at some point. Humidity is needed for heating in the instruments to avoid getting too wet. Wind info is needed for opening and closing of the hatches, etc.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: Good question and frankly I don't know. I have asked Tiago about this (he was involved with writing the ICD) and I'll let you know as soon as I have an answer.

sal_interfaces/DREAM/DREAM_Commands.xml
wvreeven: I see. The WeatherStation only provides the azimuth of the wind direction so that's all we can provide.

PR #469, opened 2021-07-01
DM-29906: Removed all commands since they are supposed to be used int…

wvreeven:
…ernally. Also removed an unused telemetry item.


pkubanek:
Internal or not, how will you open DREAM hatch? Shouldn't there still be an option for somebody at control room to close DREAM hatch (e.g. not only put it into standby)? Dropping weather command is good, but I am a bit curious about rationale to drop those commands - does that mean you will be able to manually open hatch only from DREAM EUI/GUI? My experience with M1M3 shows that making sure almost everything (except firmware programming) is available as SAL/DDS command and then build GUI/EUI on top of SAL/DDS commands helps making control distributed.

wvreeven:
Tiago and I discussed this and the idea is that the DREAM CSC will be quite autonomous. It will mostly be in ENABLED state and will determine for itself (based on WeatherStation and other telemetry) whether or the hatch should be opened or closed. Details will be worked out over the next weeks.

tribeiro (approved):

PR #470, opened 2021-07-06
DM-31017: Test: remove char and octet fields

r-owen:
Note that no other SAL components use either of these two data types.


colinwinslow (approved):
Looks good!

PR #471, opened 2021-07-14
DM-31081: MTMount: update events

r-owen:


wvreeven (changes requested):
I think there is something wrong with the MTMount ts_xml modifications that you made. You mapped all system states to powerState but for for instance the oil supply system the cooling and oil states are different in the Tekniker documentation than the power state enum. The order of the items is different because for power they are {0: OFF, 1: ON, 2: FAULT, 3: TURNING_ON, 4: TURNING_OFF} while for cooling and oil they are {0: OFF, 1: ON, 2: TURNING_ON, 3: TURNING_OFF, 4: FAULT}. Unless the documentation I am looking at is out dated.

wvreeven (approved):
It turns out that the documentation is outdated so I am approving this PR now.

PR #478, opened 2021-08-23
Add references to enums

couger01:


wvreeven (approved):
LGTM

patrickingraham:
So on the ts-xml page I see the following:
image
It's useful, but what happens most of the time is that the state returns a number, not the name. 
Is it not possible to get the number beside the name?
In the implementation done here a user is forced to click the links and go hunting.

PR #484, opened 2021-09-09
DM-31502: Added commands for starting and stopping GenericCamera auto…

wvreeven:
… exposure taking.

### DIFF
--- a/sal_interfaces/GenericCamera/GenericCamera_Commands.xml
+++ b/sal_interfaces/GenericCamera/GenericCamera_Commands.xml
@@ -115,4 +115,35 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>GenericCamera</Subsystem>
+    <EFDB_Topic>GenericCamera_command_startAutoExposure</EFDB_Topic>
+    <Description>Start automatic exposure taking</Description>
+    <item>
+      <EFDB_Name>minExpTime</EFDB_Name>
+      <Description>Minimum exposure time</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>maxExpTime</EFDB_Name>
+      <Description>Maximum exposure time</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>configuration</EFDB_Name>
+      <Description>Additional parameters in YAML format</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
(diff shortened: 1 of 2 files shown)


sal_interfaces/GenericCamera/GenericCamera_Commands.xml
```diff
@@ -115,4 +115,35 @@
...
+    </item>
+    <item>
+      <EFDB_Name>maxExpTime</EFDB_Name>
+      <Description>Maximum exposure time</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>payload</EFDB_Name>
```
tribeiro, line 137: Consider `configuration` or `config` instead of `payload`...

sal_interfaces/GenericCamera/GenericCamera_Events.xml
```diff
@@ -794,4 +794,12 @@
     <Subsystem>GenericCamera</Subsystem>
     <EFDB_Topic>GenericCamera_logevent_ccsConfigured</EFDB_Topic>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>GenericCamera</Subsystem>
+    <EFDB_Topic>GenericCamera_logevent_startAutoExposure</EFDB_Topic>
```
tribeiro, line 799: How about: `autoExposureStarted`? Also, might be a good idea to have the parameters passed to the command sent here, so the user can check that what it send is what it got.

sal_interfaces/GenericCamera/GenericCamera_Events.xml
```diff
@@ -794,4 +794,12 @@
...
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>GenericCamera</Subsystem>
+    <EFDB_Topic>GenericCamera_logevent_stopAutoExposure</EFDB_Topic>
```
tribeiro, line 803: How about `autoExposureStopped `?

tribeiro (approved):
Looks good but please, take a look at my comments before merging....

sal_interfaces/GenericCamera/GenericCamera_Events.xml
wvreeven: I modified it to `autoExposureStarted` but note that I was following the way the live view events have been implemented and there it is `startLiveView` so we may want to modify that one as well. I added the parameters that were passed on with the corresponding command.

sal_interfaces/GenericCamera/GenericCamera_Events.xml
wvreeven: I modified it to `autoExposureStopped` but note that I was following the way the live view events have been implemented and there it is `endLiveView` (not stopLiveView) so we may want to modify that one as well.

PR #485, opened 2021-09-10
DM-31715: Make LFOA event generic

tribeiro:
⚠️  By making `largeFileObjectAvailable` a generic event some components end up with no particular topic set. This causes issues with the sal `make_*` python scripts, which assumes at least one component topic set. I updated the scripts and opened lsst-ts/ts_sal#144 to support components with no particular topic set. This need to be addressed before merging this.

Add `largeFileObjectAvailable` event to the list of generics.
Remove `ATHeaderService_Events.xml` file (which only had LFOA event).
Remove `CCHeaderService_Events.xml` file (which only had LFOA event).
Remove `GCHeaderService_Events.xml` file (which only had LFOA event).
Remove `MTHeaderService_Events.xml` file (which only had LFOA event).
Remove `largeFileObjectAvailable` from `Electrometer_Events.xml`.
Remove `largeFileObjectAvailable` from `FiberSpectrograph_Events.xml`.


rbovill (approved):
Changes look good to me.  Unit tests are passing.  Approved.

mareuter (approved):
Looks good to merge when the SAL side gets squared away.

mareuter:
We need to add README files to the HeaderService directories with the explanation of why they are empty.

PR #505, opened 2021-10-29
DM-31354: Script: add checkpoint count fields and help

r-owen:


tribeiro (approved):
Looks great! thanks for the update!

PR #508, opened 2021-11-02
DM-32251: Added dew point to all humidity sensors.

wvreeven:


r-owen (approved):
Looks great. Thank you very much for adding this.

PR #510, opened 2021-11-02
DM-32429: Add fields to generic log topics

r-owen:
Add subsystem field to generic command_setLogLevel and logevent_logLevel topics.
Add timestamp field to generic logevent_logMessage topic.


mareuter (approved):
Thanks for all the changes. Looks good to me.

PR #514, opened 2021-11-08
Tickets/cap 770

tony-johnson:
Merge CAP-770 onto develop. A few notes:

* Generation of a subset of the XML files are now completely automated, including the entire Telemerty.xml and a large fraction of the Events.xml.  
* There are many places where descriptions are missing (flagged MISSING), and other small errors (such as missing units) have crept in. Rather than attempt to fix these by hand in the XML (as we have done in the past) we have created a series of JIRAs to get these fixed in the underlying CCS code, so they will be automatically and permanently fixed in future when we re-run the XML generation code after CCS has been updated.
* Due to the switch from semi-hand generated code to automatically generated code there are a large number of changes, some simply related to ordering of items in the XML. Everything is now strictly ordered so hopefully this will be the last time this happens.


mareuter (approved):
Since the unit tests are passing, I'm going to sign off on this since it's pretty hard to follow.

PR #519, opened 2021-12-15
DM-32974: ESS: add `pressure` telemetry topic

r-owen:


r-owen:
I just noticed your 2nd sentence: "Cannot use indexed CSC due to complexity of the indexed solution". The ESS CSC is, in fact, indexed. Each ESS instance (each running ESS CSC with a different SAL index) is allowed to write a given topic for more than one sensor; they just have to have unique values of sensorName. For example ESS 2 might write the temperature topic for two different multi-channel temperature measurement devices , one with sensorName = "floor" and the other with sensorName="roof".

Thus we have to be careful with these topics. Data with one value of (SAL index, sensorName) is completely unrelated to every other (SAL index, sensorName). This also means that the most recent value of a given topic (even for a given SAL index) is not a complete picture. Which in turn makes ESS-based Watcher rules a bit tricky to write; I ended up writing some special classes to handle this.

This sort of topic is unusual, but the Watcher's Alarm topic has exactly the same issue (though it uses the "name" field instead of "sensorName", to distinguish instances, and at present Watcher is not indexed -- that is likely to change if we run out of horsepower.

PR #523, opened 2021-12-28
DM-32836: change interlock.detail to interlock.enabled for MTHexapod and MTRotator

r-owen:
change MTHexapod application status to a scalar from an array


teweitsai (approved):
Looks good!

PR #525, opened 2022-01-04
DM-32671: Added new MotionStates to MTDome.

wvreeven:


sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -10,7 +10,32 @@
     MotionState_Parked,
     MotionState_Parking,
     MotionState_Stopped,
-    MotionState_Stopping
+    MotionState_Stopping,
+    MotionState_Stopping_Braking,
+    MotionState_Stopped_Braked,
+    MotionState_Brakes_Disengaged,
+    MotionState_Brakes_Engaged,
```
r-owen, line 17: The fact that you have separate stopped_braked, and brakes_engaged states is very confusing to me, since with the brakes fully engaged I would expected both of these states to mean exactly the same thing.

More generally I feel that MotionState is badly overloaded: it is being used to report things that really should be reported separately, each using its own field and its own enum:
* Motor power: on/off/turning_on/turning_off
* Motor cooling: on/off/turning_on/turning_off
* Seal inflation: inflated/deflated/inflating/deflating
* Brake State: engaged/disengaged/engaging/disengaging

That way I'd know things like: what is the state of the brakes and motor power while while inflating or deflating the seals?

Another way to express this is to "normalize" the state information, so that separate fields are used to report separate kinds of state.

r-owen (changes requested):
Please at least ditch one of stopped_braked and brakes_engaged or clearly document (in ts_idl) how they differ. But I very strongly suggest normalizing the state -- either in this ticket or another. It would be best if the vendor did this so you could just pass along their reports.

sal_interfaces/MTDome/MTDome_Events.xml
wvreeven: I explained on Slack how the two states are different using the state machine of the LabVIEW software. I also indicated that events can be published for motor power, motor cooling, seal inflation and the brake states and that I can do that in a separate ticket.

r-owen (approved):
Please consider a ticket to report motor power, seal inflation, etc. independently of motion state. It would be nice if we could use that work to simplify the motion states at the same time, but that may be too difficult.

PR #526, opened 2022-01-04
Remove master from CSC table

couger01:


r-owen (approved):
Looks good. Thanks for the improvement.

You might consider removing the `index:` prefix for further simplification.

PR #532, opened 2022-01-19
Fix not handling newline denominated enumerations

couger01:


python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
                             write_heading(cf, name="Enumerations")
                             enumeration_seen = True
                         states = topic.text.split(",")
+                        print(states)
                         state_name = states[0].split("_")[0].lstrip()
                         cf.write(f":{state_name}:\n")
                         for state in states:
+                            state = state.split("_")[1:]
+                            state = "".join(state)
```
r-owen, line 299: Please consider replacing the above two lines with:
```
state = state.split("_", 1)[1]
```
where the 2nd argument to `split` specifies the maximum number of splits.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
                             write_heading(cf, name="Enumerations")
                             enumeration_seen = True
                         states = topic.text.split(",")
+                        print(states)
                         state_name = states[0].split("_")[0].lstrip()
                         cf.write(f":{state_name}:\n")
                         for state in states:
+                            state = state.split("_")[1:]
```
r-owen, line 298: Please add a code comment explaining the format of `state`. That should make the code much easier to understand.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
                             write_heading(cf, name="Enumerations")
                             enumeration_seen = True
                         states = topic.text.split(",")
+                        print(states)
```
r-owen, line 294: Is this intentional, or a debug statement accidentally left in?

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
...
                         state_name = states[0].split("_")[0].lstrip()
                         cf.write(f":{state_name}:\n")
                         for state in states:
+                            state = state.split("_")[1:]
+                            state = "".join(state)
+                            state = state.lstrip()
```
r-owen, line 300: Is this line actually needed? This is all the text after the first underscore, so I'm puzzled as to what you can be stripping out here. (That said: the comment I requested earlier explaining the format might clarify).

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
...
+                            state = state.split("=")
+                            state = "".join(state[0])
                             cf.write(
-                                f"  * :any:`{state.split('_')[1]} <lsst.ts.idl.enums.{subsystem}.{state_name}.{state.split('_')[1].upper()}>`\n"  # noqa
+                                f"  * :any:`{state} <lsst.ts.idl.enums.{subsystem}.{state_name}.{state.upper()}>`\n"  # noqa
                             )
+                            print(state)
```
r-owen, line 306: Another leftover debug statement?

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -291,12 +291,19 @@ def generate_subsystems_doc():
...
                         states = topic.text.split(",")
+                        print(states)
                         state_name = states[0].split("_")[0].lstrip()
                         cf.write(f":{state_name}:\n")
                         for state in states:
+                            state = state.split("_")[1:]
+                            state = "".join(state)
+                            state = state.lstrip()
+                            state = state.split("=")
+                            state = "".join(state[0])
```
r-owen, line 302: I am puzzled by these two lines. It looks like you split at an equals and then feed the first item to `join`. Splitting at an equals makes sense to me, but feeding a single string to `join` does not. I am pretty sure that has to be a no-op. (I'm a bit surprised it even works, but strings are iterable).

Also please consider rstrip after the split, because there may be spaces before the `=`.

r-owen (commented):
I have a few questions in line and one bigger question here: I don't understand how this new code handles newline-delimited enums. I don't see reference to a newline anywhere.

r-owen:
Another question: are you stripping the numerical values? They'd be useful to show. In fact I'd suggest showing numeric values even when none are specified, but that might be too much work. Whereas it should be easy to show the values when they have been specified.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -294,9 +294,13 @@ def generate_subsystems_doc():
...
                         cf.write(f":{state_name}:\n")
                         for state in states:
-                            cf.write(
-                                f"  * :any:`{state.split('_')[1]} <lsst.ts.idl.enums.{subsystem}.{state_name}.{state.split('_')[1].upper()}>`\n"  # noqa
-                            )
+                            # state format starts as [\n statename_state ]
+                            state = state.split("_")[
+                                1:
+                            ]  # Some states have more than one underscore split.
+                            state = "".join(state)  # Make that list a string
```
r-owen, line 301: Please change lines 298 - 301 to:
```
state = state.split("_", 1)[1]
```
there is no need to do more than a single split, and having split at only the first "_" you don't have to re-assemble anything.

Also please consider `enum_item` instead of `state` as the variable name and "enum_entry" instead of "state" for the thing you are splitting.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -294,9 +294,13 @@ def generate_subsystems_doc():
...
+                            state = state.lstrip()  # Strip the new line
```
r-owen, line 302: Please delete this line.

If you split "\n foo_bar_baz" with `state = state.split("_", 1)[1] ` you end up with `state = "bar_baz"`. The \n is already gone because it was part of the first piece of the split that you threw away: the discarded string is `"\n foo"`.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -294,9 +294,13 @@ def generate_subsystems_doc():
                         state_name = states[0].split("_")[0].lstrip()
                         cf.write(f":{state_name}:\n")
                         for state in states:
-                            cf.write(
-                                f"  * :any:`{state.split('_')[1]} <lsst.ts.idl.enums.{subsystem}.{state_name}.{state.split('_')[1].upper()}>`\n"  # noqa
-                            )
+                            # state format starts as [\n statename_state ]
```
r-owen, line 297: At line 294 please use `state_name = states[0].split("_", 1)[0].lstrip()`

In other words: add the `1` argument to `split`. There is no point to splitting past the first "_".

r-owen: Thanks for documenting the format. However, itcan include a value. Thus please consider something along the lines of:
```
# enum item format is `\n enumClassName_itemName [= value]`
# where itemName may include underscores,
# and all whitespace is optional and of arbitrary length.
```

r-owen: Please consider using `enum_class_name` instead of `state_name` to avoid confusion. Each enum entry typically does describe a single state (though there may be exceptions). But even if so, enum class name is *never* a state.

I'm sure there are other terms. But C++ and Python both use the term "enum class", and Java uses "enum type", which is close. So "enum_class_name" is a decent choice.

r-owen (changes requested):
Please remove an unneeded join and an unneeded lstrip.

More minor optional requests include slightly better explanation of the format and slightly clearer variable names.

python/lsst/ts/xml/generate_subsystems_doc.py
```diff
@@ -294,9 +294,15 @@ def generate_subsystems_doc():
...
-                                f"  * :any:`{state.split('_')[1]} <lsst.ts.idl.enums.{subsystem}.{state_name}.{state.split('_')[1].upper()}>`\n"  # noqa
-                            )
+                            # enum item format `\n enumClassName_itemName
+                            # [=value]`
+                            # where itemName may include underscores,
+                            # and all whitespace is optional and of
+                            # arbitrary length.
+                            state = state.split("_", 1)[
+                                1
+                            ]  # only split after the first underscore.
```
r-owen, line 304: You split *at* the first underscore. I think what you are trying to say is "strip the beginning through the first underscore"

r-owen (approved):
Please clarify a code comment. Then it looks good.

PR #538, opened 2022-02-23
CAP-856: MTHexapod and MTRotator: change connected event

r-owen:
to a have a single "connected" field


teweitsai (approved):
Looks good. Thanks for this update!

PR #540, opened 2022-02-23
tickets/CAP-812

mareuter:
Remove PromptProcessing from interface.


rbovill (approved):
PromptProcessing was removed.

PR #541, opened 2022-02-24
Add missing bonn-shutter telemetty for AuxTel

tony-johnson:


sal_interfaces/ATCamera/ATCamera_Telemetry.xml
```diff
@@ -604,4 +604,24 @@
...
     </item>
   </SALTelemetry>
+  <!--Created by xmlmaker2 for subsystem ATCamera class ATCamera_bonn_shutter_Device using level 1-->
+  <!--============================================================================================-->
+  <SALTelemetry>
+    <Subsystem>ATCamera</Subsystem>
+    <EFDB_Topic>ATCamera_bonn_shutter_Device</EFDB_Topic>
+    <item>
+      <EFDB_Name>bonn_V36</EFDB_Name>
+      <Description>BonnShutter 5-volt reading</Description>
```
mareuter, line 614: I think this should say 3.6 volt reading based on the topic name.

mareuter (approved):
Please have a look at the comment, but otherwise this is fine.

sal_interfaces/ATCamera/ATCamera_Telemetry.xml
tony-johnson: Yes, it should but the fix needs to be made in the CCS subsystem which generates the XML. Will create a JIRA to get it fixed.

PR #554, opened 2022-03-18
DM-34026: Added the resetDrivesAz and calibrateAz commands.

wvreeven:


sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -150,4 +150,21 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_command_resetDrivesAz</EFDB_Topic>
+    <Description>Reset one or more AZ drives. This is necessary when exiting from FAULT state without going to Degraded Mode since the drives don't reset themselves.</Description>
+    <item>
+      <EFDB_Name>reset</EFDB_Name>
+      <Description>Desired reset action to execute on each AZ drive: 0 means don't reset, 1 means reset.</Description>
```
r-owen, line 159: I strongly suggest making the data type `boolean` instead of `float`. If you anticipate needing more values in the future then consider `integer` and an enum.

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -150,4 +150,21 @@
...
+      <IDL_Type>float</IDL_Type>
+      <Units>unitless</Units>
+      <Count>5</Count>
+    </item>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_command_calibrateAz</EFDB_Topic>
+    <Description>Take the current position of the dome as zero. This is necessary as long as the racks and pinions on the drives have not been installed yet to compensate for slippage of the drives.</Description>
```
r-owen, line 168: It is a pity to have to add a special command for this. If you intend to leave this in long-term then please consider a more descriptive name, such as zeroAz or setZeroAz or....

sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -150,4 +150,21 @@
...
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_command_resetDrivesAz</EFDB_Topic>
+    <Description>Reset one or more AZ drives. This is necessary when exiting from FAULT state without going to Degraded Mode since the drives don't reset themselves.</Description>
+    <item>
+      <EFDB_Name>reset</EFDB_Name>
+      <Description>Desired reset action to execute on each AZ drive: 0 means don't reset, 1 means reset.</Description>
+      <IDL_Type>float</IDL_Type>
```
r-owen, line 160: I strongly suggest making the data type boolean instead of float. If you anticipate needing more values in the future then consider integer and an enum.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: It is a temporary command and this will probably be removed within the next year and certainly before on sky commissioning starts. However, I like the proposal so I will change it to setZeroAz.

sal_interfaces/MTDome/MTDome_Commands.xml
wvreeven: Good idea. I'll do that.

r-owen (approved):
Looks great. Thank you for the changes.

PR #557, opened 2022-03-25
Thermal system engineering mode

pkubanek:
- added engineerMode flag
- enableEngineeringMode command


rbovill (approved):
New topics added look good to me.

PR #560, opened 2022-03-29
DM-34220: Updated developer info for DIMM, DREAM and GenericCamera.

wvreeven:


wvreeven:
It is indeed. And unfortunately the image used is managed by Rob, who is on vacation this week. So I cannot merge this yet.

PR #564, opened 2022-04-04
Implement CAP-870

tony-johnson:
Additions to support Calibration mode, see discussion at https://lsstc.slack.com/archives/CLQ63T89L/p1647278833410459 and https://jira.lsstcorp.org/browse/CAP-870

It particular this change adds two new items to startIntegration event, and adds an additional state (DISCARDING) to raftsDetailedState.

@menanteau Note that the intention is that the header service starts using the new timeout item in startIntegration to determine the header service timeout, instead of the current behavior of using expTime.

### DIFF
--- a/sal_interfaces/ATCamera/ATCamera_Events.xml
+++ b/sal_interfaces/ATCamera/ATCamera_Events.xml
@@ -493,6 +493,24 @@
       <Units>second</Units>
       <Count>1</Count>
     </item>
+    <!-- Items added for CAP-870 -->
+    <item>
+      <EFDB_Name>mode</EFDB_Name>
+      <Description>The camera data taking mode (normal/calibration/(TBD in future))</Description>
+      <IDL_Type>string</IDL_Type>
+      <IDL_Size>1</IDL_Size>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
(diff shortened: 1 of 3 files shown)


menanteau:
@tony-johnson Thanks for the heads up. The definition of the timeout is a bit more involved. At the moment the timeout for the headerservice is set as:
`timeout = metadata_exptime[timeout_keyword] + timeout_exptime`
at the moment we use:

`timeout_exptime: 30`
`timeout_keyword: 'EXPTIME'`

Where `timeout_exptime` is the time (configurable) after the exposure time to give up. We are using 30s.

Are you suggesting a wholesale change of `timeout` or `timeout_exptime`

tony-johnson:
Hi Felipe, right now I believe you do the following:

Get the value of startIntegration.exposureTime and add 30 seconds to it, and use it as the header service timeout. 

I am suggesting in future you get the value of startIntegration.timeout and add 30 seconds to it, and use it as the timeout

The reason for this is that in camera calibration mode the exposureTime is unknown at start integration time, and thus set to zero, but the value of timeout is known and will be filled in appropriately in both calibration and normal mode.

Note, there is no reason to add a TIMEOUT fits header keyword.

mareuter (approved):
XML changes look good to me.

PR #565, opened 2022-04-04
DM-31447: MTRotator: add motor current and voltage to the motors telemetry topic

r-owen:


teweitsai (approved):
Looks good. Thanks for this improvement.

PR #567, opened 2022-04-04
Tickets/cap 875

tony-johnson:
This pull request consists of 3 commits. It may be easier to review the individual commits than the total. They are:

* Add Image handling SettingsApplied events (CAP-875)
* Change version from long to string (CAP-874)
* Change all occurrences of SettingsApplied to configuration (CAP-874)


sal_interfaces/ATCamera/ATCamera_Events.xml
```diff
@@ -6464,18 +6464,189 @@
...
+      <Description>MISSING</Description>
+      <IDL_Type>long long</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>heartbeat_taskPeriodMillis</EFDB_Name>
+      <Description>MISSING</Description>
+      <IDL_Type>long long</IDL_Type>
+      <Units>unitless</Units>
```
mareuter, line 6490: Shouldn't the attributes with `Millis` in the name have units as `second`?

sal_interfaces/ATCamera/ATCamera_Events.xml
```diff
@@ -6464,18 +6464,189 @@
...
+    <item>
+      <EFDB_Name>version</EFDB_Name>
+      <Description>Version of the settings, formatted using CCS conventions</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>agentmonitorservice_taskPeriodMillis</EFDB_Name>
+      <Description>MISSING</Description>
```
mareuter, line 6481: Be nice to see all of the `MISSING` descriptions filled in at some point. I understand if it doesn't happen right now.

mareuter (approved):
One type of attribute needs a unit check. Otherwise I think this is fine.

sal_interfaces/ATCamera/ATCamera_Events.xml
tony-johnson: Yes, or I would think milliSecond. The units as well as the description comes from the CCS core, and currently they are missing there. There is a JIRA: https://jira.slac.stanford.edu/browse/LSSTCCS-2482 which is marked as resolved in core 6.0.3 but we have not yet switched to using that release.

PR #576, opened 2022-04-26
DM-34557: Uncomment commented-out ATWhiteLight enum value (now that ts_sal can handle such a large value)

r-owen:
As of [DM-34567](https://jira.lsstcorp.org/browse/DM-34567) ts_sal now uses `long long` for enumerations (after a false start adding a type attribute to the Enumeration tag that turned out to not work). Uncomment the one ATWhiteLight enum value that was too large for the old type `long`.


r-owen:
Dave decided to change how large enum values are supported. The "type" attribute is gone and the new type for all enum values will be "long long". I updated this code accordingly.

pkubanek (approved):
Approved.

PR #577, opened 2022-04-28
Remove unsigned from DIMM items

dmills-lsst:


r-owen (approved):
This looks fine to me. That said, I'm wondering why the change? ESS can handle unsigned long (uint32). It just can't handle unsigned long long (uint64).

PR #578, opened 2022-05-03
CAP-833: Delete AdamSensors

r-owen:


wvreeven:
Has it been decided to completely remove AdamSensors? It is still present in ts_cycle_build.

mareuter:
It was granted removal from the XML interface. We will remove it from the cycle build as part of cycle 26 work.

wvreeven:
Thanks for the clarification!

rbovill (approved):
Approved.

PR #580, opened 2022-05-11
DM-34771: rename "salIndex" public ScriptQueue fields to "scriptSalIndex"

r-owen:
And update the test of reserved words to catch use of "salIndex".


dmills-lsst (approved):
ok, hope the other changes do not break something else in CI tests

PR #581, opened 2022-05-12
Fix for CAP-894

tony-johnson:
Fix instances where strings had count>1 (CAP-894)

### DIFF
--- a/sal_interfaces/CCCamera/CCCamera_Events.xml
+++ b/sal_interfaces/CCCamera/CCCamera_Events.xml
@@ -7345,18 +7345,18 @@
     <!--CCSMETA: Array manSerNum indexed by location-->
     <item>
       <EFDB_Name>manSerNum</EFDB_Name>
-      <Description>MISSING</Description>
+      <Description>MISSING (colon delimited array)</Description>
       <IDL_Type>string</IDL_Type>
       <Units>unitless</Units>
-      <Count>9</Count>
+      <Count>1</Count>
     </item>
     <!--CCSMETA: Array name indexed by location-->
     <item>
       <EFDB_Name>name</EFDB_Name>
-      <Description>MISSING</Description>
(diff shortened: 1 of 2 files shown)


sal_interfaces/CCCamera/CCCamera_Events.xml
```diff
@@ -7345,18 +7345,18 @@
     <!--CCSMETA: Array manSerNum indexed by location-->
     <item>
       <EFDB_Name>manSerNum</EFDB_Name>
-      <Description>MISSING</Description>
+      <Description>MISSING (colon delimited array)</Description>
```
r-owen, line 7348: Can you please change MISSING to something more descriptive? Here for any of the other topics for which you can provide some kind of description.

I assume the data is presently colon-separated? Because if not, we use comma as the separator for everywhere except for lists of paths.

r-owen (approved):
Thank you for fixing this. Two requests, if you have time:
* Please change MISSING to an actual description of what the field contains (in many places).
* If the data is not already being output as a single colon-separated string, then please consider using a comma as the separator -- at least for non-path lists of strings. That matches our convention.

sal_interfaces/CCCamera/CCCamera_Events.xml
tony-johnson: HI Russell, these files are generated completely automatically. Where the description is MISSING this indicates that the documentation is either missing in the CCS source, or that there is an issue with the way the description is being transferred. In both cases we are working on fixes, but it will take several release cycles to get them all fixed. 

We agreed long ago to send all camera string arrays as colon delimited, with backslash used as a escape in case the strings contains a colon (or backslash). https://jira.lsstcorp.org/browse/CAP-570. I know the header service relies on this behavior so I would not like to change it.

PR #586, opened 2022-05-25
DM-34972: MTMount: add actualTorquePercentage to the cameraCableWrap telemetry topic and fix the units for the existing actualTorque fields

r-owen:


sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -90,11 +83,11 @@
...
     <item>
-      <EFDB_Name>actualAcceleration</EFDB_Name>
-      <Description>Actual acceleration.</Description>
+      <EFDB_Name>actualTorquePercentage</EFDB_Name>
+      <Description>Actual torque for each actuator, as a percentage.</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>deg/s^2</Units>
-      <Count>1</Count>
+      <Units>%</Units>
+      <Count>2</Count>
```
pkubanek, line 90: I see a bit of a naming issue. This presumably comes from motors. LBT uses rear and front, as you have those two on traditional hydrostatic bearing systems (they pull/push against each other to remove backlash).

sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -90,11 +83,11 @@
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>actualAcceleration</EFDB_Name>
-      <Description>Actual acceleration.</Description>
+      <EFDB_Name>actualTorquePercentage</EFDB_Name>
+      <Description>Actual torque for each actuator, as a percentage.</Description>
```
pkubanek, line 87: Should't be "Actual torque for each **motor**, as a percentage."?

PR #591, opened 2022-06-06
DM-35126: MTMount: add homeBothAxes command plus other tweaks

r-owen:
to match events from the TMA


wvreeven (approved):
Looks good to me.

PR #592, opened 2022-06-07
Tickets/cap 920

tony-johnson:
Updates for AuxTel and ComCam based on differences in installed CCS subsystems vs XML 11.1.
Note the default git diff does a poor job on XML, the changes are actually much less than it makes it appear.

### DIFF
--- a/sal_interfaces/CCCamera/CCCamera_Telemetry.xml
+++ b/sal_interfaces/CCCamera/CCCamera_Telemetry.xml
@@ -21,9 +21,9 @@
     </item>
     <item>
       <EFDB_Name>steppermotor_MotorTemperature</EFDB_Name>
-      <Description>motor temperature, degrees C</Description>
+      <Description>motor temperature</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>unitless</Units>
+      <Units>Celsius</Units>
       <Count>1</Count>
     </item>
   </SALTelemetry>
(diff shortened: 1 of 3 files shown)


sal_interfaces/CCCamera/CCCamera_Telemetry.xml
```diff
@@ -21,9 +21,9 @@
     </item>
     <item>
       <EFDB_Name>steppermotor_MotorTemperature</EFDB_Name>
-      <Description>motor temperature, degrees C</Description>
+      <Description>motor temperature</Description>
       <IDL_Type>double</IDL_Type>
-      <Units>unitless</Units>
+      <Units>Celsius</Units>
```
pkubanek, line 26: deg_C is official astropy unit.

mareuter: Celsius is also a valid astropy unit, otherwise the unit test we have checking units would have failed. We only require that the unit name be valid to astropy.

pkubanek: Yes. But Celsius is an alias.At least M1M3 preference is for un-aliased names: https://docs.astropy.org/en/stable/units/index.html .. which is true for the whole project except Camera:
````
[petr@pcd47720-1-a ts_xml]$ grep -rl Celsius sal_interfaces/
sal_interfaces/ATCamera/ATCamera_Telemetry.xml
sal_interfaces/ATCamera/ATCamera_Events.xml
sal_interfaces/GenericCamera/GenericCamera_Telemetry.xml
sal_interfaces/MTCamera/MTCamera_Telemetry.xml
sal_interfaces/MTCamera/MTCamera_Events.xml
sal_interfaces/CCCamera/CCCamera_Telemetry.xml
sal_interfaces/CCCamera/CCCamera_Events.xml
```

mareuter: Uniformity and validity are two different things. If you wish to convince Tony to change, you are free to do so. Again, the only requirement we have is that the unit value produce a valid astropy unit quantity.

tonyj321: Will consider this for a future update

tonyj321 (approved):
Have made changes suggested by @mareuter so proceeding to approve

rbovill (approved):
LGtM

PR #608, opened 2022-08-03
DM-35056: turn some events into telemetry

pkubanek:
- M1M3 Thermal System auxiliary devices
- remove aberration forces
- moved events into telemetry


rbovill (approved):
I scanned the topics and things seem to be in order.  All the unit tests are passing, so it looks good to me.  If any issues arise, they will be dealt with separately.

PR #609, opened 2022-08-03
DM-35663: remove most unused DIMM events and telemetry

r-owen:
This change is backward compatible because ts_dimm does not write any of those topics.
We should also remove the commands, but that has to be carefully scheduled, because it is not backwards compatible.


dmills-lsst:
I think we want to keep logevent_moduleStatus and logevent_dimmData, all the data may not be there yet but it will be

r-owen:
I think moduleStatus needs a rewrite to be useful. All the states are boolean, with no explanation of what true or false means.

I will put dimmData back, but I don't see how to get most of those fields from the Astelco DIMM.

dmills-lsst (approved):
ok

PR #611, opened 2022-08-10
Add main camera filter changer configuration and telemetry (fist appr…

tony-johnson:
This pull requests adds a first approximation for the telemetry and configuration to be published for the full camera filter changer. Although this will almost certainly change, and has many missing descriptions, I think it does pass all the tests, and getting it pushed to the develop branch so it will be included in the generated jar files would be very useful for testing. 

There are no changes to ATCamera or CCCamera XML


rbovill (approved):
I scanned a few items and didn't see anything amiss.  All the unit tests are passing.  LGtM.

PR #614, opened 2022-08-15
Remove duplicate names

tony-johnson:
Same as before but with duplicate names:

loader_LFD
loader_LFS

removed.


rbovill (approved):
Duplicate entries removed.

PR #617, opened 2022-09-08
Renamed the 'searchZeroShutter' command to 'homeShutter'.

wvreeven:


tribeiro:
I think I asked you this in slack but I don't remember the answer. Are you sure the Shutter is the only thing that will require homing? If not, I would suggest naming this command `home`, `homeAxis` or `homeSubsystem` and adding an attribute with an "id", "axis" or "subsystem" attribute. Then we would be able to do:

```
await mtdome.cmd_home.set_start(id=SubsystemId.APSCS)
```


and we would be able the expand the command later without the need to touch the xml.

tribeiro (approved):
Looks good!

PR #618, opened 2022-09-16
Define DREAM CSC XML - telemetry, events and commands.

pkubanek:


rbovill (approved):
All unit tests are passing.  LGtM.  Approved.

PR #624, opened 2022-10-17
DM-36648: generalize ESS topics

r-owen:


sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -14,43 +14,50 @@
...
+      <Description>Time at which the most recent data was read (TAI unix seconds).</Description>
       <IDL_Type>double</IDL_Type>
       <Units>second</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>ux</EFDB_Name>
-      <Description>x-axis wind speed.</Description>
+      <EFDB_Name>speed</EFDB_Name>
+      <Description>Median wind speed (x, y, z) measured over a period on the order of a minute.</Description>
```
wvreeven, line 24: I'd prefer that the period over which the speed is reported is determined by the configuration of the CSC. If you agree then please amend this Description and other similar ones below accordingly.

sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -14,43 +14,50 @@
...
+      <Description>Maximum wind speed magnitude measured over a period on the order of a minute.</Description>
       <IDL_Type>float</IDL_Type>
       <Units>m/s</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>ts</EFDB_Name>
-      <Description>Sonic temperature (air temperature measured sonically). This has poor absolute accuracy (it can be off by several degrees) but good time resolution.</Description>
+      <EFDB_Name>sonicTemperature</EFDB_Name>
+      <Description>Sonic temperature (air temperature measured sonically); median of values measured over a period on the order of a minute. This has poor absolute accuracy (it can be off by several degrees) but good time resolution.</Description>
```
wvreeven, line 52: Please remove " but has good time resolution" because that doesn't apply anymore, since we're collecting the median of values now.

sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -14,43 +14,50 @@
...
       <IDL_Type>float</IDL_Type>
       <Units>deg_C</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>status</EFDB_Name>
-      <Description>Device status. The details depend on the device, but 0 always means that no problems have been detected.</Description>
-      <IDL_Type>short</IDL_Type>
+      <EFDB_Name>isok</EFDB_Name>
+      <Description>Is the data valid?</Description>
```
wvreeven, line 59: How is this different than the "sensorStatus" event? It feels redundant to have this "isok" item here and for all other sensors where this has been added.

sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -74,29 +81,89 @@
...
-      <Description>Time at which the data was determined (TAI unix seconds).</Description>
+      <Description>Time at which the most recent data was read (TAI unix seconds).</Description>
       <IDL_Type>double</IDL_Type>
       <Units>second</Units>
       <Count>1</Count>
     </item>
     <item>
       <EFDB_Name>direction</EFDB_Name>
-      <Description>Wind direction: 0 = north, 90 = east.</Description>
+      <Description>Median (mean for some sensors) wind direction measured over a period on the order of a minute: 0 = north, 90 = east.</Description>
```
wvreeven, line 91: We may have discussed this already and if we did I apologize for not having remembered. Why median in some and mean in other cases? Is it possible or even favorable to always report the median or always the mean instead? Same question for the other items that report either median or mean below.

wvreeven (changes requested):
I am going to push back on several changes you made. I am available for a chat if you want.

sal_interfaces/ESS/ESS_Events.xml
```diff
@@ -60,4 +60,65 @@
...
+    </item>
+    <item>
+      <EFDB_Name>sensorStatus</EFDB_Name>
+      <Description>Sensor status. The details depend on the sensor, but 0 always means that no problems have been detected.</Description>
+      <IDL_Type>int</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>serverStatus</EFDB_Name>
```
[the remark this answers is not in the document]
r-owen: sensorStatus is status reported by the sensor, if available. Not all sensors report status.

serverStatus reports status of the data server or perhaps the connection to the data server. One example is the response_code field that RPiDataClient sends to its methods that process telemetry.

sal_interfaces/ESS/ESS_Telemetry.xml
r-owen: The period will be a combination of the sensor configuration (data reporting interval) and CSC configuration (how many samples to aggregate). I'm not sure how to say that concisely in the XML. Any suggestions?

sal_interfaces/ESS/ESS_Telemetry.xml
r-owen: Good point. However, I do want to tell users that the point of sonic temperature is that it has good time resolution. I propose to reword it as follows:

Median sonic temperature (air temperature measured sonically). Sonic temperature has poor absolute accuracy (it can be off by several degrees) but good time resolution.

and also add sonicTemperatureStdDev, to give a better picture of the fluctuations.

sal_interfaces/ESS/ESS_Telemetry.xml
r-owen: Median is a "best effort". We can compute it for sonic anemometers, but for weather stations we are stuck with whatever they report -- presumably mean.

wvreeven (approved):
Looks good to me. Thanks for the constructive discussion about these changes.

PR #625, opened 2022-10-18
Scheduler: Add commands and events to handle new block feature and ev…

tribeiro:
…ent to publish the observing mode information.


sal_interfaces/Scheduler/Scheduler_Commands.xml
```diff
@@ -36,4 +36,72 @@
...
+      <EFDB_Name>id</EFDB_Name>
+      <Description>The name of the block (e.g., OBS-123).</Description>
+      <IDL_Type>string</IDL_Type>
+      <IDL_Size>1</IDL_Size>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>override</EFDB_Name>
+      <Description>Override information in the block with those provided here.</Description>
```
patrickingraham, line 53: Issue here is that it doesn't say what the input is... which is the JSON string..
Maybe something like:
"Override block with a new JSON string provided here"  ?
I don't like that much either honestly....

sal_interfaces/Scheduler/Scheduler_Commands.xml
```diff
@@ -36,4 +36,72 @@
...
+      <IDL_Type>string</IDL_Type>
+      <IDL_Size>1</IDL_Size>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>Scheduler</Subsystem>
+    <EFDB_Topic>Scheduler_command_validateBlock</EFDB_Topic>
+    <Description>Validate block.</Description>
```
patrickingraham, line 76: All blocks get validated when the CSC gets enabled right?
If so, then then the `id` parameter is useless.

How does the CSC see new blocks? Presumably it has to go to Standby and back up, yes?

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
       <Count>1000</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>Scheduler</Subsystem>
+    <EFDB_Topic>Scheduler_logevent_observingMode</EFDB_Topic>
```
patrickingraham, line 1182: What is this? I don't think this was in the document we wrote, no?

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+    <item>
+      <EFDB_Name>id</EFDB_Name>
+      <Description>Id of the block (e.g. OBS-123).</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>statusId</EFDB_Name>
+      <Description>Id of the status.</Description>
```
patrickingraham, line 1226: Maybe, "Id of the status as an enumeration" ?

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+    <item>
+      <EFDB_Name>status</EFDB_Name>
+      <Description>Status of the block, as a human readable string.</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>iterationsCompleted</EFDB_Name>
+      <Description>Number of completed iterations.</Description>
```
patrickingraham, line 1240: A block gets executed...
should this be `executionsCompleted` ?
Then the descriptions can be 
`Number of successful executions"  ?

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+      <IDL_Type>long</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>iterationsTotal</EFDB_Name>
```
patrickingraham, line 1246: same deal as above if you make that change

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+      <Description>Total number of iterations.</Description>
```
patrickingraham, line 1247: Consider:
Total number of iterations (or executions) to be completed"

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+    </item>
+    <item>
+      <EFDB_Name>hash</EFDB_Name>
+      <Description>A unique identifier for this block.</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>configuration</EFDB_Name>
```
patrickingraham, line 1260: What is this for? What do you mean by configurations of the block?

sal_interfaces/Scheduler/Scheduler_Events.xml
```diff
@@ -1177,4 +1177,110 @@
...
+    <Description>Inventory of all the available blocks.</Description>
+    <item>
+      <EFDB_Name>ids</EFDB_Name>
+      <Description>Comma-separated list of block ids.</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>status</EFDB_Name>
```
patrickingraham, line 1279: I was expecting that there would be 1 parameter per status type.
How would your format this as a single string?

OBS-123 (Ready), OBS-234 (completed) etc?  That'd be challenging to parse.

patrickingraham (changes requested):

sal_interfaces/Scheduler/Scheduler_Commands.xml
tribeiro: You can ask to validate a particular block with an override.

I think the CSC can monitor the block directory and validate things when they change.

sal_interfaces/Scheduler/Scheduler_Events.xml
tribeiro: This is something else that LOVE needs.. but it is related to the change we are about to do.. It will basically tell what the Scheduler is doing at the moment (e.g. executing a user provided block)

sal_interfaces/Scheduler/Scheduler_Events.xml
tribeiro: I was thinking this would contain the block definition.. it will basically have the block json string.. if there are no overrides, it will be exactly the same as the file, if there are overrides, they will show here.

sal_interfaces/Scheduler/Scheduler_Events.xml
tribeiro: The ids would be in the [ids](https://github.com/lsst-ts/ts_xml/blob/2a137cc26a532c572147b68118ea3682712735bd/sal_interfaces/Scheduler/Scheduler_Events.xml#L1272) attribute and the status here, e.g.;

```
ids="OBS-123,OBS-234"
status="ready,completed"
```

The problem of having keyword for the status is that we can't update them if we realized we missed something..

sal_interfaces/Scheduler/Scheduler_Commands.xml
tribeiro: I would rather not tie in with a particular solution (e.g. JSON) here.. Just in case we want to change to some other format later. In the CSC codebase I will document what should be the format.

sal_interfaces/Scheduler/Scheduler_Commands.xml
patrickingraham: ok, not exactly sure how that works as the block directory will be a database. So you think it'll continually poll the database for all tickets marked as "Ready" and have not been validated? 
We can leave this in.. i'm just struggling to see how it fits in the big picture.

sal_interfaces/Scheduler/Scheduler_Events.xml
patrickingraham: Can you use "definition" instead? 
I'm worried that configuration is already overloaded with SAL Script and CSC configs...

sal_interfaces/Scheduler/Scheduler_Events.xml
patrickingraham: fair, but if there are 20 obs IDs and you care about the status of the 10th one then it'll be tough...
Maybe we leave it like this but change it once we think we have the status' correct?

tribeiro: Hopefully there will be a view in LOVE to parse this information into a nice format.. but certainly, once we have a better understanding about how we use this, we can change it up..

patrickingraham (approved):
Thanks! Excited to get this implemented!

PR #626, opened 2022-10-19
Update FCS telemetry and events to level 1

tony-johnson:
See if this makes the main camera filter changer XML compilable by SAL


rbovill (approved):
All tests and checks are passing.  Approved, pending CI build.

PR #628, opened 2022-10-20
DM-36189 (and DM-36672): Add prereq_jobs and output_dataset_types parameters.

ktlim:
prereq_jobs allows a comma-separated list of job ids to be provided; this job will not start until those jobs have completed. (DM-31689)

output_dataset_types allows a comma-separated list of glob patterns specifying dataset types to be provided.  Datasets matching those types will be captured by the UWS back-end system, and URIs pointing to those datasets will be included in the job result event. (DM-36672)


mareuter (approved):
Looks good to me.

PR #629, opened 2022-10-21
Backout CAP-933 changes, revert to versions from Aug 10

tony-johnson:
Backout CAP-933 changes for xml 13 release -- pending fixes to SAL.


rbovill (approved):
LGtM

PR #636, opened 2022-11-02
Tickets/dm36802 - Convert ts_xml to pyproject.toml

rbovill:
* Moved sal_interfaces/ and schema/ directories under python/lsst/ts/xml/data
* Updated Python modules and unit tests, accordingly.
* Added pyproject.toml, removed requirements.txt and updated setup.cfg and conda/meta.yaml.


conda/meta.yaml
```diff
@@ -0,0 +1,36 @@
+{% set data= load_setup_py_data() %}
+package: 
+  name: ts-integrationtests
```
tribeiro, line 3: This should probably be `ts-xml`

conda/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  version: {{ data.get('version') }}
+
+source:
+  path: ..
+
+build:
+  noarch: generic
+  script: python -m pip install --ignore-installed -e .
```
tribeiro, line 11: Take a look at [this confluence page](https://confluence.lsstcorp.org/x/0Il-Cw), it has some updates to the conda recipe to make them forward compatible with different versions of python..

python/lsst/ts/xml/generate_csv_table.py
```diff
@@ -32,7 +32,7 @@ def generate_csv_table():
     tree = ElementTree.parse(sal_interfaces_dir / "SALSubsystems.xml")
     root = tree.getroot()
 
-    outpath = utils.get_pkg_root() / "doc" / "subsystemData.csv"
+    outpath = utils.get_pkg_root() / "subsystemData.csv"
```
wvreeven, line 35: I still very strongly think that the `subsystemData.csv` file should not be moved from the `doc` directory. Are you sure that this is absolutely necessary? I don't think that any changes to the `generate_csv_table.py` file are necessary, including the change to line 141.

wvreeven (changes requested):
Sorry for pushing back on the changes made to one python file. I really don't think those are necessary and I fear they will still break the way the Makefile in the doc directory has been implemented.

r-owen (approved):
I have one suggested change and I'll leave it to you and @wvreeven to decide what to do about that one generated file.

setup.cfg
```diff
@@ -1,18 +1,9 @@
...
 max-line-length = 110
 max-doc-length = 79
 ignore = E133, E226, E228, N802, N803, N806, N812, N813, N815, N816, W503
 exclude =
-  sal_interfaces/*,
-  schema/*,
-  scripts/*,
-  tests/.tests,
+  doc,
+  doc/conf.py,
```
r-owen, line 7: I gently suggest making doc/conf.py pass flake8. It makes more a much more pleasant experience when editing the file in some editors. It does require some # noqa entries. Here is an example from ts_utils:
```
from documenteer.conf.pipelinespkg import *  # type: ignore # noqa
import lsst.ts.utils  # noqa

project = "ts_utils"
html_theme_options["logotext"] = project  # type: ignore # noqa
html_title = project
html_short_title = project
# Avoid warning: Could not find tag file _doxygen/doxygen.tag
doxylink = {}  # type: ignore
```

r-owen (approved):
Looks fine to me

conda/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+source:
+  path: ..
+
+build:
+  noarch: generic
+  script: {{ PYTHON }} -m pip install --no-deps --ignore-installed .
+
+test:
+  requires:
+    - astropy
```
tribeiro, line 15: Please, include 

```
    - ts-conda-build =0.3
```

here.

conda/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - python {{ python }}
+    - pip
+    - setuptools_scm
+    - setuptools
+  build:
+    - python {{ python }}
+    - setuptools_scm
+    - setuptools
+  run:
+    - python {{ python }}
```
tribeiro, line 36: I think you also need `astropy` and `lxml` as run dependencies (but no `ts-conda-build`, which is only a test dependency).

tribeiro (approved):
Looks good! I did left a few more comments on the conda recipe! please, take a look before merging. The current recipe won't build properly because it is missing a test dependency and it won't run properly because it is missing some run dependencies.

wvreeven (approved):
Thanks for the additional changes. Looks good to me now.

PR #638, opened 2022-11-03
Fix up schema location and order of XML elements for new ts_xml conve…

tony-johnson:
Fix up schema location and order of XML elements for new ts_xml conventions


rbovill (approved):
LGtM

PR #643, opened 2022-11-10
Improved ESS lightning and high electric field alert events. Added ES…

wvreeven:
…S lightning and electric field telemetry.

### DIFF
--- a/python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
@@ -4,59 +4,52 @@
   <SALEvent>
     <Subsystem>ESS</Subsystem>
     <EFDB_Topic>ESS_logevent_lightningStrike</EFDB_Topic>
-    <Description>Lightning strikes and high electrical field warnings.</Description>
+    <Description>Lightning strike, if detected. If all values are zero then this indicates that no strike was detected for a configurable amount of time since the last strike.</Description>
     <item>
       <EFDB_Name>sensorName</EFDB_Name>
-      <Description>The name of the sensor.</Description>
+      <Description>The user-defined name of the sensor (e.g. AuxTel LD-250).</Description>
       <IDL_Type>string</IDL_Type>
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>strike</EFDB_Name>
-      <Description>Lighting strike alarms, as a mask with numStrikeBits bits. 1 means an alarm is active. Higher-order bits are for strikes that are closer, and thus more worrisome. Each alarm will eventually reset itself if no further activity in that alarm range is detected.</Description>
-      <IDL_Type>unsigned short</IDL_Type>
-      <IDL_Size>1</IDL_Size>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
-    <item>
-      <EFDB_Name>highField</EFDB_Name>
-      <Description>High electrical field alarms, as a mask with numHighFieldBits bits. 1 means an alarm is active. Higher-order bits are for larger electrical fields, and thus are more worrisome. Each alarm will eventually reset itself if no further activity in that alarm range is detected</Description>
-      <IDL_Type>unsigned short</IDL_Type>
-      <IDL_Size>1</IDL_Size>
-      <Units>unitless</Units>
-      <Count>1</Count>
-    </item>
-    <item>
-      <EFDB_Name>numStrikeBits</EFDB_Name>
-      <Description>The maximum number of strike bits.</Description>
-      <IDL_Type>unsigned short</IDL_Type>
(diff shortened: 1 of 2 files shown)


python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
```diff
@@ -1,61 +1,69 @@
...
+      <Units>degree</Units>
       <Count>1</Count>
     </item>
     <item>
-      <EFDB_Name>numHighFieldBits</EFDB_Name>
-      <Description>The maximum number of high field bits.</Description>
-      <IDL_Type>unsigned short</IDL_Type>
-      <IDL_Size>1</IDL_Size>
+      <EFDB_Name>level</EFDB_Name>
+      <Description>Alarm level, an AlarmLevel enum limited to Red, Orange and Yellow.</Description>
```
r-owen, line 39: I hope you plan to indicate "no strike" in this event when the ESS starts up and some "reasonable" time after the most recent strike is detected. (The lightning system consolidating box does this and it's generally a good idea).

If you can have more than one strike event active at the same time then you probably need some kind of ID field so we can tell the difference between each strike event.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
```diff
@@ -1,61 +1,69 @@
...
+      <IDL_Type>int</IDL_Type>
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>ESS</Subsystem>
+    <EFDB_Topic>ESS_logevent_highElectricField</EFDB_Topic>
+    <Description>High electric field, if exceeding an intensity threshold for a certain amount of time. The thresholds for Red, Orange and Yellow and the time limit can be configured in ts_config_ocs.</Description>
```
r-owen, line 48: Same comment as strike: do you plan to time this out after a time? Can there be more than one "high field" event active at the same time (in which case you may need to add some kind of ID field).

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
r-owen: I wonder if you should bother with a level? It seems like we could determine our own severity directly from the data -- the Watcher will probably have a rule to handle this. Or does this level come from the sensor?

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -494,4 +494,82 @@
...
+    <item>
+      <EFDB_Name>timestamp</EFDB_Name>
+      <Description>Starting time of data acqusition (TAI Unix seconds) used to measure EFSs.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>strength</EFDB_Name>
+      <Description>Strength of the electric field (range -20.00 to 20.00).</Description>
```
r-owen, line 516: Please consider publishing max, mean or median, and std dev or quartiles.

If you want to just publish the one value please explain what it is -- max? mean?

I suggest having the measure in the name, e.g. `strengthMax` (and `strengthStdDev` if you add them)

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -494,4 +494,82 @@
       <Count>1</Count>
     </item>
   </SALTelemetry>
+  <SALTelemetry>
+    <EFDB_Topic>ESS_electricFieldStrength</EFDB_Topic>
+    <Description>Electric field strength for a period of 10 seconds (20 measurements per second).</Description>
```
r-owen, line 499: I suggest deleting  "(20 measurements per second)" because it may not be true for all sensors and it may be configurable for some.

I worry that delaying this data makes it less useful, so I'm glad you have the high field event as well as this telemetry.

I strongly suggest make it 1/second instead of 1/10 seconds so it has more value as a warning. Or make the timespan configurable, in which case change the "for a period of 10 seconds" to something like "over a configurable period".

r-owen (approved):
Looks very good. A few minor suggestions.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
wvreeven: Good points all of them. I have removed the enum plus the items referring to them and adjusted the descriptions.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Events.xml
wvreeven: See my previous comment.

PR #648, opened 2022-11-24
Add trigger command and events to TunableLaser

couger01:


python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
```diff
@@ -36,6 +36,10 @@
     <Subsystem>TunableLaser</Subsystem>
     <EFDB_Topic>TunableLaser_logevent_continuousModeSet</EFDB_Topic>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_logevent_triggerSet</EFDB_Topic>
```
r-owen, line 41: Please add a Description (if you keep it).

Also the name sounds like a mode to me -- something that can be set or unset. Please consider `triggered` or `burstTriggered`.

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
 <SALEventSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://github.com/lsst-ts/ts_xml/blob/develop/schema/SALEventSet.xsd">
-  <Enumeration>DetailedState_NonPropagatingState,DetailedState_PropagatingState</Enumeration>
+  <Enumeration>DetailedState_NonPropagatingState,DetailedState_PropagatingState,DetailedState_PropagatingTriggerState</Enumeration>
```
r-owen, line 4: This new state seems to duplicate your new triggerSet event. Please fix that. Here are two possible solutions:

1) Keep the new `triggerSet` event and change the new detailed state to something like `BurstMode`, meaning the laser is armed and might be triggered at any time.

   This reduces traffic. When you trigger the laser you only have to output a single "triggered" event. This may be a bit easier for users to work with. For example the user can simply wait for that event to know that the laser has been triggered.


2) Ditch the new `triggerSet` event and describe burst mode using *two* new detailed state events (instead of the one you added), e.g.: `BurstModeWaitingForTrigger` and `BurstModeTriggered`.

   This generates more traffic (2 events per trigger: BurstModeTriggered followed immediately by BurstModeWaitingForTrigger), which may be a bit messier for users. But it does have the advantage that the detailed state tells you *everything* about whether the laser is propagating. The user doesn't have to know about a separate "triggered" event.

r-owen (changes requested):

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
```diff
@@ -43,4 +43,8 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_command_trigger</EFDB_Topic>
```
r-owen, line 48: Please consider `triggerBurst` for clarity.

Please add a Description.

r-owen: Also please consider combining the setBurstMode and setBurstCount commands into a single setBurstMode command with a `burstCount` field. That would give the user more predictable behavior -- when they enable burst mode they know exactly how many pulses they will get per trigger.

r-owen (approved):
Looks great. Thanks for consolidating the count.

PR #649, opened 2022-11-29
DM-37113: MTMount_Events: fix cameraCableWrapControllerSettings

r-owen:
The limits enabled fields did not match the information available from the TMA.


couger01 (approved):
LGTM.

PR #664, opened 2023-02-07
tickets/DM-37860: Removed all references to ATArchiver and CCArchiver; those CSCs were …

rbovill:
…removed.


r-owen (approved):
It looks to me as if you got all instances.

One question: can you also get rid of CatchupArchiver?

rbovill:
> 

It seems that the CatchupArchiver is still a necessary CSC.

PR #666, opened 2023-02-13
Made sure that the lightning strike bearing and mean/median/standard …

wvreeven:
…deviation are floats instead of shorts.


wvreeven:
No, the sensor reports them as int values.

pkubanek (approved):

PR #669, opened 2023-02-23
Fix M1M3 description strings

pkubanek:
- Fixed Description strings
- total force/moment


rbovill (approved):
Descriptions were fixed and updated.

PR #674, opened 2023-03-01
Tickets/dm 38162

tribeiro:


r-owen (approved):

tribeiro:
I used the wrong branch name.. this should have been DM-38060 and not DM-38162.

python/lsst/ts/xml/data/sal_interfaces/MTPtg/MTPtg_Events.xml
```diff
@@ -812,4 +812,37 @@
...
+    <item>
+      <EFDB_Name>height</EFDB_Name>
+      <Description>Site height above see level.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>meter</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>timeZone</EFDB_Name>
+      <Description>Time zone offset.</Description>
```
pkubanek, line 842: What about summer time? timeZone changes every 6 months in Chile.

r-owen: If we have to update this when clocks change then we should consider just deleting this field (from both pointing components). It's not very interesting (especially compared to the other 3 fields, which also have the advantage of being static by any reasonable definition). I suppose the description could be clearer.

PR #676, opened 2023-03-06
Tickets/dm 38061

wvreeven:


python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -585,5 +592,51 @@
...
+      <IDL_Type>double</IDL_Type>
+      <Units>second</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>frequency</EFDB_Name>
+      <Description>The frequency of the spectrum trace.</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>Hz</Units>
+      <Count>751</Count>
```
r-owen, line 625: If the frequency is spaced in some predictable way, can you publish a few model coefficients, instead? For example if the frequency is evenly spaces, then min and interval (or min and max) would suffice.

It would dramatically reduce the amount of data being stored.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -585,5 +592,51 @@
...
+    </item>
+    <item>
+      <EFDB_Name>trace</EFDB_Name>
```
r-owen, line 628: I gently suggest `amplitude` (if it is one) or perhaps `spectrum`.

r-owen (commented):
I'm reluctant to approve this until I get feedback on the "frequency" field.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -585,5 +592,51 @@
...
+    <Description>Spectrum analyzer trace (once per second).</Description>
+    <item>
+      <EFDB_Name>sensorName</EFDB_Name>
+      <Description>The user-defined name of the sensor (e.g. AuxTel SSA3000X).</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>timestamp</EFDB_Name>
```
r-owen, line 614: I suspect this timestamp field is unnecessary, though I see that topics have it and I'm a but uncomfortable with that. You get the private_sndStamp field for free, and unless the system is slow to send the data around, it's probably just as good. In this specific case I think it is as good.

I'm not sure if we should do a global purge, have it even when not needed, or have a mix of "timestamp is present when it adds useful information, absent otherwise".

Despite the name, private_sndStamp is not really private -- all languages can see it.

wvreeven: I would prefer not to have a mix. Either a global purge or leave it as is. For now I have chosen to leave as is because that introduces no code changes to ts_ess_csc.

pkubanek: I will need global purge. M1M3 is overusing timestamp (I know when I need it for resolution, but anyway 50Hz aren't so bad..). Can we set a date? It's a mess. I will probably need way to set private_sndStamp in C code (as cRIOs FPGAs are running internal clocks, and it will be better to supply private_sndStamp value to that), probably shall start a ticket?

wvreeven: We might as well try to get this done now. There will be a ts_xml release in a few weeks so why not make use of that?

r-owen: @pkubanek if you feel you have to set private_sndStamp in your own code then you probably need a separate `timestamp` field. private_sndStamp *must* be the time at which the message is sent. That way we can measure latency in transporting messages.

For many topics (such as ESS) that time is close enough to the time at which the data was measured that we can live with the difference. But we also have plenty of topics for which that is not the case. In particular most of the telemetry from MTMount, MTRotator, and ATMCS needs and has a separate `timestamp` field.

PR #683, opened 2023-04-03
DM-38545: Script and ScriptQueue: add instrument field

r-owen:
Add instrument field to the Script metadata event
and to the ScriptQueue nextVisit event.


tribeiro (approved):
Looks good!

PR #684, opened 2023-04-03
DM-38542: ESS: simplify acceleration and accelerationPSD

r-owen:
telemetry topics:

* accelerationPSD:

  * change array length from 200 to 201. This gives nicer frequencies for round values of maxPSDFrequency.
  * eliminate interval: it was irrelevant.
  * eliminate  minPSDFrequency: it is now always 0.
  * eliminate numDataPoints: it is now always 201 (all array elements).

* acceleration:

  * change array length from 200 to 400. This allows publishing acceleration once for each time we publish acccelerationPSD.


couger01 (approved):
LGTM.

PR #686, opened 2023-04-04
DM-35122: Pointing sends bad frame value when using MTCS.point_azel.

tribeiro:
* Add new coordinate frames for azel, planet and ephemeris targets.


mareuter (approved):
Looks good to me.

PR #687, opened 2023-04-04
DM-38558: use ts_pre_commit_conf

r-owen:


rbovill (approved):
It all looks good to me.

PR #693, opened 2023-04-17
DM-38724: ATWhiteLight: lampState: add lightDetected and ignitionFailures

r-owen:
fields and return LampControllerState_StandbyOrOn, instead of separate Standby and On values.


wvreeven (approved):
Looks good to me.

PR #716, opened 2023-07-01
enable/disable Force Components command

pkubanek:


doc/version-history.rst
```diff
@@ -4,6 +4,11 @@
 Version History
 ===============
 
+v18.0.0
```
tribeiro, line 7: The next scheduled release is v17.1.0 below. I think you should just add your change to the existing items, no need to create a new release.

python/lsst/ts/xml/data/sal_interfaces/MTM1M3/MTM1M3_Commands.xml
```diff
@@ -787,4 +787,37 @@
...
+        active engineering mode. Switching from the "active engineering mode"
+        will revert applied flags back to their default values.
+    </Description>
+    <item>
+      <EFDB_Name>forceComponent</EFDB_Name>
+      <Description>Force component to Enable or Disable. One of the EnableDisable_ enumeration member.</Description>
+      <IDL_Type>short</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+      <Enumeration>
```
tribeiro, line 804: I think enumerations are defined at the top of the [event file](https://github.com/lsst-ts/ts_xml/blob/d5fe4a55c28edb579e156d7d2e89e7b243633f6e/python/lsst/ts/xml/data/sal_interfaces/MTM1M3/MTM1M3_Events.xml#L3), here you should just add something like [this](https://github.com/lsst-ts/ts_xml/blob/d5fe4a55c28edb579e156d7d2e89e7b243633f6e/python/lsst/ts/xml/data/sal_interfaces/MTPtg/MTPtg_Events.xml#L65).

tribeiro (commented):
Can you add an event that specify which corrections are enabled no?

Also, See my inline comment about how we specify enumeration.

python/lsst/ts/xml/data/sal_interfaces/MTM1M3/MTM1M3_Commands.xml
pkubanek: Well, that top placement doesn't pass through C++ salobject generator. Looks like a bug in ts_sal then, will verify that.

tribeiro: weird, how to you get the enums for DetailedState?

doc/version-history.rst
pkubanek: Fixed. Thanks.

rbovill (approved):
LGtM

python/lsst/ts/xml/data/sal_interfaces/MTM1M3/MTM1M3_Commands.xml
pkubanek: I cannot move it to global commands enumeration, as then I lost shared/global event enumerations. Has to keep it that way now.

pkubanek:
> Can you add an event that specify which corrections are enabled no?

That's already there, in ForceActuatorState. I plan to move that out to something more descriptive, but that can wait for another PR - let's keep it simple for this one.

PR #733, opened 2023-08-09
DM-40220: Add command in the MTAOS to apply offset to the optical state of the system.

tribeiro:


python/lsst/ts/xml/data/sal_interfaces/MTAOS/MTAOS_Commands.xml
```diff
@@ -160,4 +160,21 @@
...
   </SALCommand>
+  <SALCommand>
+    <Subsystem>MTAOS</Subsystem>
+    <EFDB_Topic>MTAOS_command_offsetDOF</EFDB_Topic>
+    <Description>Utility command to apply offsets to the degrees of freedom (DOF). Units will match that of the degreeOfFreedom event.</Description>
+    <item>
+      <EFDB_Name>value</EFDB_Name>
+      <Description>Array with the offsets to be applied to the DOF.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>micron</Units>
```
teweitsai, line 171: This part how about we use the `unitless` instead and put the following to the description: "Please check the file of  state0inDof.yaml in ts_ofc/policy"?

The benefit is that the user will know the meaning of each DOF.

Otherwise, it would be very confusing because the 50 DOFs have the units of `um` and  `arcsec`.

In addition, because the OFC will accumulate the internal DOF at each iteration, you may want to identify this command will overwrite the current accumulated DOF or accumulate with it? This question applies to the `MTAOS_command_resetOffsetDOF` as well.

python/lsst/ts/xml/data/sal_interfaces/MTAOS/MTAOS_Events.xml
```diff
@@ -59,6 +59,13 @@
       <Units>micron</Units>
       <Count>50</Count>
     </item>
+    <item>
+      <EFDB_Name>userDoF</EFDB_Name>
+      <Description>User applied offsets.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>micron</Units>
```
teweitsai, line 66: The same question of unit as above. I will ask the user to check the policy file in ts_ofc to know the definition and unit for each DOF.

python/lsst/ts/xml/data/sal_interfaces/MTAOS/MTAOS_Telemetry.xml
```diff
@@ -0,0 +1,28 @@
...
+    <Subsystem>MTAOS</Subsystem>
+    <EFDB_Topic>MTAOS_m2BendingModes</EFDB_Topic>
+    <Description>Measured M2 bending modes values.</Description>
+    <item>
+      <EFDB_Name>value</EFDB_Name>
+      <Description>Amplitude of the measured bending mode.</Description>
+      <IDL_Type>double</IDL_Type>
+      <Units>micron</Units>
+      <Count>20</Count>
+    </item>
```
teweitsai, line 26: How do we realized this in code? The unit of M1M3 and M2 is the Newton for each actuator, not the bending mode in um. We could estimate I guess, but not measure it directly for sure.

teweitsai (approved):
Looks good.

PR #740, opened 2023-08-23
Add ``SummaryState`` enumeration to ``SALGenerics``.

tribeiro:


pkubanek (approved):
I suggest those changes:

```
diff --git a/doc/version-history.rst b/doc/version-history.rst
index 198ba36..b407ea5 100644
--- a/doc/version-history.rst
+++ b/doc/version-history.rst
@@ -8,7 +8,7 @@ v20.0.0
 -------
 
 * Copy enumerations for ts-idl into a new enums submodule.
-* Allow components to still define SummaryState enumerations in their xml files while generic enumerations are not supported by SAL.
+* Allow components to still define SummaryState enumerations in their xml files while generic enumerations are not supported by C/C++ SAL.
 * Move the code that defines SAL topics structure and generate avro-schema files from the kafka version of salobj.
   * Add private_revCode back to the generic fields.
 * Update enumerations to match the definitions from the enums submodule (see interface updates).
@@ -65,10 +65,6 @@ v20.0.0
 
     * Add ErrorCode enumeration.
 
-  * MTVMS
-    
-    * Remove unnecessary summaryState enumeration.
-
   * TunableLaser
 
     * Replace detailedState enumeration with LaserDetailedState.
```

PR #745, opened 2023-09-05
Tickets/dm 40580

tribeiro:


tests/test_SALSubsystems.py
```diff
@@ -10,7 +10,10 @@
 
 INDEX_ENUM_CHECK = re.compile(r"[^,= \w]+")
 
-VALID_LANGUAGES = {"CPP", "Java", "LabVIEW", "IDL", "Python"}
+# These are the "languages" we need to build libraries for. Note we removed
+# support for python runtime (previously SALPY) and the python/salobj
+# components will either use IDL (for DDS) or xml directly (for kafka).
```
couger01, line 15: I think you should clarify if the runtime would be blank if kafka or remain IDL for both.

couger01 (approved):
LGTM, just one clarification desired.

tests/test_SALSubsystems.py
tribeiro: that is a good point.. however, I don't think we are ready to make that call

tests/test_SALSubsystems.py
```diff
@@ -10,7 +10,10 @@
...
+VALID_RUNTIME_LANGUAGES = {"CPP", "Java", "LabVIEW", "IDL"}
```
wvreeven, line 16: Should LabVIEW still be mentioned here? It looks like you removed it everywhere from SALSubsystems.xml

wvreeven (approved):
Looks good to me. I left one comment for your consideration.

tests/test_SALSubsystems.py
couger01: Maybe remove the "or xml directly (for kafka)" since that could be potentially confusing for the context of choosing a runtime language.

tests/test_SALSubsystems.py
tribeiro: no, Labview is still quoted in a couple systems. Especially:

- MTMount (probably not needed any longer)
- ATMCS (still needed)
- ATPneumatics (still needed)
- MTM2 (probably not needed any longer)
- Test (no comments)

We may deprecated it when we go to kafka.

tests/test_SALSubsystems.py
tribeiro: ok

PR #747, opened 2023-09-11
DM-36101: In kafka salobj remove protections against interfering with EFD ingest of DDS-based SAL

tribeiro:


mareuter (approved):
Looks good to me. The doc builds failed though.

PR #760, opened 2023-10-04
Initial commit of LEDProjector xml

JosephParsons:
for the LED options, it should return 3 lists that all sync up.
One list being the numerical index of the LED, (1-10)
the second list being the serial number itself, I.e. "M505L4"
The third list being the actual data we want, suchas is the LED on true/false
That way we have like such:
[1, 2, ..., 10]
["M505L4", "M375L4", ... ""]
[true, false, ...., false]


python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,33 @@
...
+    <Description>Turn on all LEDs connected to the labjack.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_turnAllLEDsOff</EFDB_Topic>
+    <Description>Turn off all LEDs connected to the labjack.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_turnOnLED</EFDB_Topic>
```
tribeiro, line 16: I'd suggest following the same pattern you used for the previous 2 commands. Consider renaming this `turnLEDOn`, or you could drop "LED" and call it `turnOn`. However, and being extra picky here, consider using `switch` instead  of `turn`.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsl"?>
+<SALCommandSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALCommandSet.xsd">
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_turnAllLEDsOn</EFDB_Topic>
```
tribeiro, line 6: Consider dropping the "LED" from the command name, also, consider using "switch" instead, e.g. `LEDProjector_command_switchAllOn`.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,33 @@
...
+    <Description>Turn on all LEDs connected to the labjack.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_turnAllLEDsOff</EFDB_Topic>
```
tribeiro, line 11: Same as above, consider `switchAllOff`.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,33 @@
...
+    <Description>Turn off all LEDs connected to the labjack.</Description>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_turnOnLED</EFDB_Topic>
+    <Description>Turn on specific LED, named by serial number or numerical ID.</Description>
+    <item>
+      <EFDB_Name>LED_By_SerialNumber</EFDB_Name>
```
tribeiro, line 19: Consider `serialNumber` only.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,33 @@
...
+      <Description>Serial number of LED to turn on.</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+    <item>
+      <EFDB_Name>LED_By_Index</EFDB_Name>
```
tribeiro, line 26: consider `index` only.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+    LabjackControllerState_Cooldown=2,
+    LabjackControllerState_Error=3
+  </Enumeration>
+  <SALEvent>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_logevent_labjackConnected</EFDB_Topic>
+    <Description>Is the CSC connected to the labjack controller interfacing with the LEDs?</Description>
+    <item>
+      <EFDB_Name>connected</EFDB_Name>
+      <Description>Connected if true, else disconnected</Description>
```
tribeiro, line 23: Consider "true if connected, false otherwise."

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+  </Enumeration>
+  <Enumeration>
+    LabjackControllerState_Unknown=0,
+    LabjackControllerState_StandbyOrOn=1,
+    LabjackControllerState_Cooldown=2,
+    LabjackControllerState_Error=3
+  </Enumeration>
+  <SALEvent>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_logevent_labjackConnected</EFDB_Topic>
```
tribeiro, line 19: Consider merging this event with `labjackState` below. Basically, add `connected` to the event. However, you could also create a new `Disconnected` state or assume `Unknown` means disconnected.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_logevent_ledState</EFDB_Topic>
+    <Description>LED status list for all LEDs connected</Description>
+    <item>
+      <EFDB_Name>numericalID</EFDB_Name>
```
tribeiro, line 46: I'd suggest shortening this to `id` only.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+      <EFDB_Name>ledState</EFDB_Name>
+      <Description>LED state; a LEDBasicState enum.</Description>
+      <IDL_Type>int</IDL_Type>
+      <Units>unitless</Units>
+      <Count>10</Count>
+    </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_logevent_ledOnMinutes</EFDB_Topic>
```
tribeiro, line 69: Consider renaming this event to something like `ledDutyCycle`. 

When do you imagine this event will be published? Something like, when the user sends an on/off command? I imagine you don't want to be publishing this constantly right?

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+    <Description>List of estimated times the leds were on in minutes</Description>
+    <item>
+      <EFDB_Name>numericalID</EFDB_Name>
```
tribeiro, line 72: consider shortening this to `id` only.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,93 @@
...
+    </item>
+    <item>
+      <EFDB_Name>serialNumber</EFDB_Name>
+      <Description>List of LED serial-numbers correlating with list data</Description>
+      <IDL_Type>string</IDL_Type>
+      <Units>unitless</Units>
+      <Count>10</Count>
+    </item>
+    <item>
+      <EFDB_Name>minutes</EFDB_Name>
```
tribeiro, line 86: Consider something like `onTime`.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Telemetry.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsl"?>
+<SALTelemetrySet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALTelemetrySet.xsd">
+  <SALTelemetry>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_LEDStates</EFDB_Topic>
```
tribeiro, line 6: Usually telemetry topics are something you are constantly monitor and publishing. Things like temperature etc. A state is probably something that will change asynchronously so, there's no need to constantly publish it.

tribeiro (commented):
This looks pretty good! 

Please take a look at my inline comments. 

Also, a couple of general comments, 

I see in `LEDProjector_logevent_ledState` and in other topics you created attributes with 10 elements. Although this is a perfectly valid approach, there are a couple things you need to consider. 

1. This will tie you to only being able to support up to a fixed number of LEDs (in this case 10).  Maybe that is all you will ever had but it is a pretty strong constraint you are imposing to the system.
2. We do not support vector of string. The way you specified `serialNumber` in `LEDProjector_logevent_ledState` means that this attribute is a string with 10 characters and not an array of 10 strings. 

My suggestion is that you make this events belong to each individual LED. Basically something like this:

```
  <item>
      <EFDB_Name>id</EFDB_Name>
      <Description>LED numerical index.</Description>
      <IDL_Type>int</IDL_Type>
      <Units>unitless</Units>
      <Count>1</Count>
    </item>
    <item>
      <EFDB_Name>serialNumber</EFDB_Name>
      <Description>LED serial-number.</Description>
      <IDL_Type>string</IDL_Type>
      <Units>unitless</Units>
      <Count>1</Count>  <!-- This is a string with unlimited size -->
    </item>
    <item>
      <EFDB_Name>ledState</EFDB_Name>
      <Description>LED state; a LEDBasicState enum.</Description>
      <IDL_Type>int</IDL_Type>
      <Units>unitless</Units>
      <Count>1</Count>
    </item>
```

Then you would publish the individual states of each LED. You will need to be careful not to update them all at once or data could be overwritten in the EFD. But this would allow you to adapt to the number of LEDs.

doc/version-history.rst
```diff
@@ -3,6 +3,18 @@
 ===============
 Version History
 ===============
+v21.0.0
+-------
+
+* Interface updates:
```
tribeiro, line 9: Please, add these changes to version 20.0.0 below.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
```diff
@@ -0,0 +1,78 @@
...
+      <EFDB_Name>index</EFDB_Name>
+      <Description>Index number of LED to switch off.</Description>
+      <IDL_Type>int</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_command_elapsedTimeOn</EFDB_Topic>
```
tribeiro, line 54: Not sure I understand what this command is supposed to do. Is this a way to query for how long a lamp has been on? Or is it a way to switch on a lamp for a specified amount of time?

Commands are like actions the CSC can take.

If this is supposed to be a way to query the up time, consider something like `LEDProjector_command_showUpTime`, and that should have `index` and `serialNumber` only, I guess. Then, you can have an event that is published as a result of this command with the information you want (e.g. `LEDProjector_logevent_upTime`, with `index`, `serialNumber` and `upTime` as parameters).

tribeiro (commented):
Alright! this is rapidly improving. A couple more things for you to consider.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Commands.xml
JosephParsons: It was supposed to show uptime, but I'm sure this can live elsewhere, so I'll just remove it.

doc/version-history.rst
```diff
@@ -66,6 +65,11 @@ v20.0.0
 
     * Add AlignComponent enumeration.
 
+  * LEDProjector
+
+    * Add turnAllLEDsOn, turnAllLEDsOff, turnOnLED, turnOffLED.
+    * Add labjackConnect, labjackState, ledStates, and ledOnMinutes events.
```
tribeiro, line 71: Please, update the list of events included to match the xml.

doc/version-history.rst
```diff
@@ -3,7 +3,6 @@
 ===============
 Version History
 ===============
-
```
tribeiro, line 6: I think you might have removed one too many empty line... 😄

tribeiro (commented):
Alright, I am ready to approve the PR. I just have a few more comments and also, I would like @parfa30 to take a look and make sure she's ok with this.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
+<SALEventSet xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsd">
+  <Enumeration>
+    LEDBasicState_Unknown=0,
+    LEDBasicState_Off=1,
+    LEDBasicState_On=2,
+    LEDBasicState_UnexpectedlyOn=3,
```
parfa30, line 8: What does UnexpectedlyOn/Off mean? What would you see to report that state?

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
```diff
@@ -0,0 +1,41 @@
...
+    LEDBasicState_UnexpectedlyOff=4,
+    LEDBasicState_Error=5,
+    LEDBasicState_LabjackError=6,
+    LEDBasicState_LabjackConnected=7,
+    LEDBasicState_LabjackDisconnected=8
+  </Enumeration>
+  <SALEvent>
+    <Subsystem>LEDProjector</Subsystem>
+    <EFDB_Topic>LEDProjector_logevent_ledState</EFDB_Topic>
```
parfa30, line 17: Would this publish as a dictionary? Or would you feed it an ID to get state?

parfa30 (approved):
I think this looks good. Just a couple questions. Main question is, would the state of every LED be published in the EFD? Or will it be published only when the state changes? If the states are published upon configuration and then published whenever changed, that's sufficient.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
JosephParsons: I kept it from ATWhiteLight. I suspect when there is a mismatch of states where the Python thinks we are on, or off, and the labjack reports that we are actually in the opposite state. I'm not sure if its needed but I figured better to have it and not need it and we can remove later.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
JosephParsons: I was thinking it would be published on state change. In addition I was thinking it would also publish labjack status change.
So if the labjack gets connected, it would be something like {0, "Labjack", 7}

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
parfa30: Makes sense.

python/lsst/ts/xml/data/sal_interfaces/LEDProjector/LEDProjector_Events.xml
parfa30: Works for me.

tribeiro (changes requested):
Please, add an entry for `LEDProjector` in the [SALSubsystems](https://github.com/lsst-ts/ts_xml/blob/develop/python/lsst/ts/xml/data/sal_interfaces/SALSubsystems.xml) file. 

It should look like this:

```
  <SALSubsystem>
    <Name>LEDProjector</Name>
    <Description> ADD SOME DESCRIPTION HERE</Description>
    <IndexEnumeration>no</IndexEnumeration>
    <AddedGenerics>csc, configurable</AddedGenerics>
    <ActiveDevelopers>Joseph Parsons</ActiveDevelopers>
    <Github>https://github.com/lsst-ts/ts_ledprojector</Github>
    <JenkinsTestResults>https://tssw-ci.lsst.org/job/LSST_Telescope-and-Site/job/ts_ledprojector/</JenkinsTestResults>
    <RubinObsContact>Parker Fagrelius</RubinObsContact>
    <CSCDocs/>
    <ProductOwner>Parker Fagrelius</ProductOwner>
    <RelatedDocuments></RelatedDocuments>
    <SoftwareLanguage>Python</SoftwareLanguage>
    <RuntimeLanguages>IDL</RuntimeLanguages>
    <VendorContact>Astronomical Consulting and Equipment</VendorContact>
    <Simulator>Internal to CSC</Simulator>
    <Configuration>https://github.com/lsst-ts/ts_config_mtcalsys</Configuration>
  </SALSubsystem>
```

I suspect once you do that one unit test will start complaining that you didn't defined the enumerations. This means you need to create a new submodule in `lsst/ts/xml/enums/LEDProjector.py` and define enumerations that match the definition in the xml file.

PR #764, opened 2023-10-05
Tickets/dm 40878

teweitsai:
* Add the new item ``copleyFaultStatus`` in ``MTRotator_electrical`` topic.
* Rename the item ``offlineSubstate`` to ``faultSubstate`` in ``MTRotator_logevent_controllerState`` topic.
* Add the new item ``drivesEnabled`` to ``MTRotator_logevent_configuration`` topic.

### DIFF
--- a/doc/version-history.rst
+++ b/doc/version-history.rst
@@ -17,6 +17,11 @@ v20.1.0
     * Rename the item ``offlineSubstate`` to ``faultSubstate`` in ``MTRotator_logevent_controllerState`` topic.
     * Add the new item ``drivesEnabled`` to ``MTRotator_logevent_configuration`` topic.
 
+  * MTHexapod
+
+    * Fix and improve the description in ``MTHexapod_actuators`` topic.
+    * Add the new item ``copleyFaultStatus`` and improve the description in ``MTHexapod_electrical`` topic.
+
 v20.0.0
 -------
 
--- a/doc/version-history.rst
+++ b/doc/version-history.rst
@@ -8,6 +8,14 @@ v20.1.0
 -------
 
 * Updated the contents of the README.
+* Interface updates:
+
+  * MTRotator
+
+    * Add FaultSubstate enumeration.
+    * Add the new item ``copleyFaultStatus`` in ``MTRotator_electrical`` topic.
+    * Rename the item ``offlineSubstate`` to ``faultSubstate`` in ``MTRotator_logevent_controllerState`` topic.
+    * Add the new item ``drivesEnabled`` to ``MTRotator_logevent_configuration`` topic.
 
 v20.0.0
 -------
--- a/python/lsst/ts/xml/data/sal_interfaces/MTRotator/MTRotator_Events.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/MTRotator/MTRotator_Events.xml
@@ -22,6 +22,11 @@
     EnabledSubstate_ConstantVelocity=6
 </Enumeration>
   <Enumeration>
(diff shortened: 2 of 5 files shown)


python/lsst/ts/xml/enums/MTRotator.py
```diff
@@ -71,11 +76,24 @@ class EnabledSubstate(enum.IntEnum):
...
 
+class FaultSubstate(enum.IntEnum):
+    """Controller substate for the FAULT state.
+
+    Value reported in ``telemetry.fault_substate``.
+
+    This is enum ``FaultSubStates`` in Simulink model.
+    """
+
+    EMERGENCY_STOPPING = 0
```
tribeiro, line 92: Are you sure you want `EMERGENCY_STOPPING` to be state with value 0? 

I think I would expect something like `UNKNOWN` or maybe ever `NO_ERROR` as the lowest enum value.

python/lsst/ts/xml/data/sal_interfaces/MTRotator/MTRotator_Events.xml
```diff
@@ -20,6 +20,11 @@
     EnabledSubstate_Initializing=4,
     EnabledSubstate_Relative=5,
     EnabledSubstate_ConstantVelocity=6
+</Enumeration>
+  <Enumeration>
+    FaultSubstate_EmergencyStopping=0,
```
tribeiro, line 25: Are you sure about having the lowest state be EmergencyStopping? this seems a bit weird to me. 

I'd suggest adding an `Unknown` state with value 0 then switch order an have `NoError=1`, `WaitClearError=2` and `EmergencyStopping=3`.

tribeiro (commented):
This looks good. However I have a couple comments I hope you will consider before I can approve the PR.

python/lsst/ts/xml/enums/MTRotator.py
teweitsai: This is to be consistent with the `FaultSubStates` added by Julen:
https://github.com/lsst-ts/ts_mt_hexRot_simulink/blob/develop/commonScripts/FaultSubStates.m

I think it might be better  to keep the same.

python/lsst/ts/xml/data/sal_interfaces/MTRotator/MTRotator_Events.xml
teweitsai: This is to be consistent with the FaultSubStates added by Julen:
https://github.com/lsst-ts/ts_mt_hexRot_simulink/blob/develop/commonScripts/FaultSubStates.m

I think it might be better to keep the same.

doc/version-history.rst
```diff
@@ -4,6 +4,18 @@
 Version History
 ===============
 
+v21.0.0
```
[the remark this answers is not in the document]
teweitsai: Fixed in the commit: [57735bc](https://github.com/lsst-ts/ts_xml/pull/764/commits/57735bc2a46885bbdbbbac174a8e9a4a587660a9). Thanks!

python/lsst/ts/xml/data/sal_interfaces/MTRotator/MTRotator_Events.xml
tribeiro: Ok, can we maybe ask them to change their state enumeration then?

tribeiro (approved):
Ok looks good. Let's see if we can change that enumeration on Tekniker's side.

PR #771, opened 2023-11-10
DM-41590: Fix type SAL-long type to AVRO-int

tribeiro:
Changes involves:

- Fix type mapping SAL-long -> AVRO-int
- Add support for floating point values to be set as ``None``.
- Write a file with the list of revcodes.
- Copy the component xml files alongside the avro schema files and also generate the generics xml file.

### DIFF
--- a/tests/test_get_component_info.py
+++ b/tests/test_get_component_info.py
@@ -32,8 +32,14 @@ from lsst.ts.xml.component_info import ComponentInfo
 class GetComponentInfoTestCase(unittest.TestCase):
     def check_result(self, result_path: pathlib.Path, name: str) -> None:
         result: dict[str, dict[str, typing.Any]] = dict(topics=dict())
+        component_result_path = result_path / name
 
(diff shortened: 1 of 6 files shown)


tests/test_get_component_info.py
```diff
@@ -45,6 +51,14 @@ def check_result(self, result_path: pathlib.Path, name: str) -> None:
                     avro_schema = json.loads(fp.read())
                     result["topics"][avro_schema["name"]] = avro_schema
 
+        assert hash_table_created
+
+        xml_type_expected = {"Commands", "Events", "Generics"}
```
mareuter, line 56: Is telemetry handled outside this?

tribeiro: The problem is that not all components have telemetry.. :(  Script, which is used for testing does not.

mareuter: OK. Thanks for the explanation.

mareuter (approved):
Looks good to me.

PR #775, opened 2023-11-15
tickets/dm41703: Enable Imperial units and add rpm and gallon/min to astropy.units instance.

rbovill:


mareuter (approved):
Looks good to me.

PR #779, opened 2023-11-27
Tickets/dm-41678

kfanning:
Adds support for tunablelaser changing its alignment.

This branch adds 1 event (`alignmentChanged`) for when the optical alignment of the laser is changed and adds 1 function definition (`changeAlignment`) for when changing the optical alignment of the laser. 

This blocks tickets DM-40029 and DM-41426 which implement the function  and use the event in the tunablelaser CSC.

### DIFF
--- a/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
@@ -26,6 +26,18 @@
   </SALCommand>
   <SALCommand>
     <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_command_setOpticalConfiguration</EFDB_Topic>
+    <item>
+      <EFDB_Name>configuration</EFDB_Name>
+      <Description>Optical alignment configuration to be set.</Description>
+      <IDL_Type>string</IDL_Type>
+      <IDL_Size>1</IDL_Size>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALCommand>
+  <SALCommand>
+    <Subsystem>TunableLaser</Subsystem>
     <EFDB_Topic>TunableLaser_command_setBurstMode</EFDB_Topic>
     <item>
       <EFDB_Name>count</EFDB_Name>
--- a/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
@@ -36,6 +36,18 @@ LaserErrorCode_HW_CPU_ERROR = 7304
   </SALEvent>
   <SALEvent>
     <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_logevent_opticalConfiguration</EFDB_Topic>
+    <item>
+      <EFDB_Name>configuration</EFDB_Name>
+      <Description>The new optical alignment configuration.</Description>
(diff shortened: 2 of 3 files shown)


python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
```diff
@@ -24,6 +24,18 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_command_changeAlignment</EFDB_Topic>
```
tribeiro, line 29: Not sure this command name has the right connotation. How about `setOpticalConfiguration`?

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
```diff
@@ -24,6 +24,18 @@
...
+    <item>
+      <EFDB_Name>optical_configuration</EFDB_Name>
```
tribeiro, line 31: topic attributes must be `lowerCamelCase` (yes, I know it is confusing, sorry). Also, I would rename this to something shorter and maybe more expressive? maybe just `configuration`?

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
```diff
@@ -34,6 +34,18 @@ LaserErrorCode_HW_CPU_ERROR = 7304
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>TunableLaser</Subsystem>
+    <EFDB_Topic>TunableLaser_logevent_alignmentChanged</EFDB_Topic>
```
tribeiro, line 39: This name gives the impression of a boolean (did the alignment changed or not?). Maybe replace this with, `opticalConfiguration`?

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml
```diff
@@ -34,6 +34,18 @@ LaserErrorCode_HW_CPU_ERROR = 7304
...
+    <item>
+      <EFDB_Name>optical_configuration</EFDB_Name>
```
tribeiro, line 41: Same as above, topic attributes should be `lowerCamelCase`.. Also, consider renaming this just `configuration` this way, you have a command `setOpticalConfiguration` with a `configuration` attribute and an `opticalConfiguration` event with `configuration` attribute. It would all match up.

tribeiro (changes requested):
Please, take a look at the comments. Also, please, rebase to develop and update the version history file with a description of the change.

python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Commands.xml
pkubanek: Shouldn't we rather use standard configuration sets? Make CSC configurable, and use the start command with configurationOverride to set the requested configuration set?

kfanning: I'm not 100% sure, but to only change optical configuration during the start command would necessitate cycling the CSC in the middle of a calibration script. I don't think this is ideal, especially since cycling the CSC will generate warnings (that the observers do need to be hearing).

tribeiro (approved):
Looks good!

PR #781, opened 2023-12-14
Adding three commands to tunable laser for the thermal sensor

JosephParsons:
Adding three commands for the temperature control thermal sensor on the laser.

1. Changing the temperature set point
2. Turning it on
3. Turning it off

### DIFF
--- a/doc/version-history.rst
+++ b/doc/version-history.rst
@@ -20,8 +20,7 @@ v21.0.0
 
   * MTM2
 
-    * Improve the description of ``MTM2_forceErrorTangent`` topic.
-
+    * Improve the description of ``MTM2_forceErrorTangent`` topic.     
 
   * Added command for mask rotation for ts_CBP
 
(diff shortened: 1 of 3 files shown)


parfa30:
As part of this ticket, could we add some Events as well? Or perhaps you want to do that in a different ticket? The events would be:

- Set point changed
- Thermal system is On/Off
- Laser fan turned ON (this is the fan that will be triggered on with the RPi)
- 100Hz sound triggers interlock (this will also be triggered on the RPi)

I wonder also if we can read the PV on the Omron as part of telemetry?

JosephParsons:
> As part of this ticket, could we add some Events as well? Or perhaps you want to do that in a different ticket? The events would be:
> 
>     * Set point changed
> 
>     * Thermal system is On/Off
> 
>     * Laser fan turned ON (this is the fan that will be triggered on with the RPi)
> 
>     * 100Hz sound triggers interlock (this will also be triggered on the RPi)
> 
> 
> I wonder also if we can read the PV on the Omron as part of telemetry?

I never commented on this.
I added the setpoint and thermal system. The RPii related stuff I put in a subtask of the RPii ticket here:
https://jira.lsstcorp.org/browse/DM-42237

parfa30 (commented):
This all looks good to me! @couger01, will you check that there are no issues with the xml structure and approve?

couger01 (approved):
LGTM.

PR #783, opened 2024-01-10
In ``field_info.py``, remove support for null type for float and double.

tribeiro:
This is creating some issues with the rollout of the C++ library so we will remove it for now and implement it again in the future.


tribeiro:
We spoke about the implications of this change and decided not to pursue it.

PR #803, opened 2024-03-04
Add logicTimerStarted and logicTimerEnded events to Electrometer

couger01:


parfa30:
So these events are used by the logic timer to start or end its measurement, so let's change these to logicTimerStart and logicTimerEnd.

parfa30 (changes requested):
change to logicTimerStart and logicTimerEnd since these events are used to tell the logic timer to start or stop recording.

parfa30 (approved):
looks good!

PR #804, opened 2024-03-07
ESS: Add telemetry for the Q330 earthquake monitor.

wvreeven:


python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -625,4 +625,372 @@
...
+    </item>
+    <item>
+      <EFDB_Name>n</EFDB_Name>
+      <Description>The north-south component.</Description>
+      <IDL_Type>float</IDL_Type>
+      <Units>m/s2</Units>
+      <Count>40</Count>
+    </item>
+    <item>
+      <EFDB_Name>z</EFDB_Name>
```
tribeiro, line 660: Consider `accelerationZ`

tribeiro (commented):
General comment. I think the name selected for the topics attributes could be improved. See my inline comment on the initial occurrences but please consider renaming these attributes to something more expressive. I suggest something like `accelerationEastWest` instead of `e`, `accelerationNothSout` instead of `n` and `accelerationZ` instead of `z`.

python/lsst/ts/xml/data/sal_interfaces/ESS/ESS_Telemetry.xml
```diff
@@ -625,4 +625,372 @@
       <Count>1</Count>
     </item>
   </SALTelemetry>
+  <SALTelemetry>
+    <EFDB_Topic>ESS_earthquakeBH</EFDB_Topic>
```
tribeiro, line 629: I just realized what the names of these topics means. I would suggest you also expand these names. It would make much easier for us to find the information in the EFD. Consider something like `ESS_earthquakeBroadBandHighGain` and so on.

tribeiro (approved):
Thank you so much! looks great!

PR #806, opened 2024-03-14
Implement CAP-1026 (pathfinder telemetry and events)

tony-johnson:
This is to allow us to send pathfinder telemetry and events as CCCamera telemetry, at the same time as the camera is on level 3 and publishing equivalent MTCamera telemetry.


rbovill (approved):
Tests are passing.  Briefly scanned the interface updates and all looks ok to me.

PR #817, opened 2024-03-22
Emergency fix for Cycle 36 (copy of fix for Cycle 35). CAP-1031

tony-johnson:
Emergency fix for Cycle 36 (copy of fix for Cycle 35). CAP-1031


rbovill (commented):
LGtM.

PR #823, opened 2024-04-10
Tickets/dm 43676

wvreeven:


python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Commands.xml
```diff
@@ -215,4 +215,16 @@
       <Count>1</Count>
     </item>
   </SALCommand>
+  <SALCommand>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_command_setPowerManagementMode</EFDB_Topic>
+    <Description>Set the power management mode.</Description>
+    <item>
+      <EFDB_Name>powerManagementMode</EFDB_Name>
```
tribeiro, line 223: consider renaming this field to `mode` only. The command is already "setPowerManagmentMode" there is no need to repeat "powerManagement" here.

python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -287,6 +293,7 @@
   <SALEvent>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_logevent_brakesEngaged</EFDB_Topic>
+    <Description>Brakes engaged.</Description>
```
tribeiro, line 296: Consider making the description a bit more descriptive. We can certainly get the "brakes engaged" part from the topic name. Consider something like:

```
Reports whether the brakes are engaged or not.
```

python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -299,6 +306,7 @@
   <SALEvent>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_logevent_lockingPinsEngaged</EFDB_Topic>
+    <Description>Locking pins engaged.</Description>
```
tribeiro, line 309: Same as above.

python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -311,6 +319,7 @@
   <SALEvent>
     <Subsystem>MTDome</Subsystem>
     <EFDB_Topic>MTDome_logevent_interlocks</EFDB_Topic>
+    <Description>Interlocks latched.</Description>
```
tribeiro, line 322: Same as above.

python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -346,4 +355,16 @@
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_logevent_powerManagementMode</EFDB_Topic>
+    <Description>Power management mode set.</Description>
+    <item>
+      <EFDB_Name>powerManagementMode</EFDB_Name>
```
tribeiro, line 363: Consider renaming to `mode` only.

python/lsst/ts/xml/data/sal_interfaces/MTDome/MTDome_Events.xml
```diff
@@ -346,4 +355,16 @@
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>MTDome</Subsystem>
+    <EFDB_Topic>MTDome_logevent_powerManagementMode</EFDB_Topic>
+    <Description>Power management mode set.</Description>
```
tribeiro, line 361: Consider something like:

> Information about the power management mode.

python/lsst/ts/xml/enums/MTDome.py
```diff
@@ -168,6 +169,15 @@ class RadLockingPinState(enum.IntEnum):
     DISENGAGED = 2
 
 
+class PowerManagementMode(enum.IntEnum):
+    """Power management mode."""
+
+    NO_POWER_MANAGEMENT = 0
```
tribeiro, line 175: I think you should probably start at 1 instead of 0. Note that, by default, enumerations start at 1. So the way you defined it in the xml `NO_POWER_MANAGEMENT` is supposed to be 1. 

If you really want this to start at zero (which I advise against), you will need to explicitly set the values in the xml, e.g.;

```
  <Enumeration>
    PowerManagementMode_NO_POWER_MANAGEMENT=0,
    PowerManagementMode_OPERATIONS=1,
    PowerManagementMode_EMERGENCY=2,
    PowerManagementMode_MAINTENANCE=3
  </Enumeration>
```

tribeiro (approved):
Looks good but please, see my comments before merging.

I hope you will consider at least renaming the topics attribute to reduce naming redundancy. Also, the enumeration needs attention.

python/lsst/ts/xml/enums/MTDome.py
wvreeven: Done. Note that I renumbered the RadLockingPinState enum values as well since they too started at 0.

PR #835, opened 2024-04-15
Add guide to documenteer dependency grouping

couger01:
The pipelines config that is available now imports from the guide
config. This means that the config expects to have any guide
dependency available but the documenteer pipeline group does not
have guide's group dependencies listed. So if we have only pipelines
listed, we don't get the guide's dependencies and thus missing
package errors start to appear. The solution is to add the
guide group to our install call which will catch any missing
packages in the future. Also simplified package install to one line
to save an unnecessary install step.


rbovill (approved):
LGtM

PR #837, opened 2024-05-13
adding command for led projector to adjust power if using DAC channel

JosephParsons:


parfa30 (commented):
Can we set all of the LEDs at the same time? And what number is the serial number?

JosephParsons:
@parfa30 I only did singular LED implementation, but I can change it to multiple channel if you'd like.
Serial number would be like "M375L4"

parfa30:
Ok, the serial number is good. I think we will want to be able to set all the LEDs at the same time and also set just one at a time. Basically, feed a dictionary.

JosephParsons:
@parfa30  hows that?
I can add a third that allows for a specific selection too, but wasn't sure if you wanted that, or just an all/single option.

parfa30 (approved):
LGTM

PR #846, opened 2024-05-17
updating tunable laser states

parfa30:
Associated with https://github.com/lsst-ts/ts_TunableLaser/pull/52

### DIFF
--- a/doc/news/interface_changes/DM-44083.tunablelaser.rst
+++ b/doc/news/interface_changes/DM-44083.tunablelaser.rst
@@ -1,0 +1,1 @@
+Changed state names for TunableLaser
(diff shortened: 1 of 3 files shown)


doc/news/DM-44083.misc.rst
couger01, line 1: Needs to be under interface_changes folder and using the name DM-44083.tunablelaser.rst.

couger01 (changes requested):
You need to update the enumeration here as well https://github.com/lsst-ts/ts_xml/blob/6b20a718d370e90034a02b34f6624f41a3a88269/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml#L4. Also you need to correct the name of the news fragment but otherwise looks good.

doc/news/DM-44083.misc.rst
parfa30: I hadn't realized that. Thanks! I moved and renamed the file.

couger01: It seems like the old file wasn't removed, could you remove this one please?

couger01 (changes requested):
Can you remove the old news fragment otherwise looks real good to me.

parfa30:
Oh shoot! I thought it was removed. It should be now.

couger01 (approved):
LGTM, make sure to rebase on latest develop and squash the commits please.

PR #858, opened 2024-05-22
Add missing Description entries

pkubanek:
develop

### DIFF
--- a/doc/news/DM-43815.mtaircompressor
+++ b/doc/news/DM-43815.mtaircompressor
@@ -1,0 +1,1 @@
+Added missing description entries to XML.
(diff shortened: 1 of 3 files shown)


doc/news/DM-43815.mtaircompressor
```diff
@@ -0,0 +1 @@
+Added missign description entries to XML.
```
wvreeven, line 1: `missign` -> `missing`. Also please always use an active voice so `Add` instead of `Added`.

wvreeven (approved):
LGTM. Please see my comment and correct the typos before merging.

PR #865, opened 2024-05-23
ATSpectrograph: Add topics description and remove unnecessary topic attributes to empty topics.

tribeiro:


mareuter (approved):
Looks good to me.

PR #875, opened 2024-07-08
tickets/DM-44757

elanaku:
Edited xml to take a group ID.

### DIFF
--- a/python/lsst/ts/xml/data/sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
@@ -35,6 +35,13 @@
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
+    <item>
+      <EFDB_Name>groupId</EFDB_Name>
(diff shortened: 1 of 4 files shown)


python/lsst/ts/xml/data/sal_interfaces/FiberSpectrograph/FiberSpectrograph_Commands.xml
```diff
@@ -35,6 +35,13 @@
       <Units>unitless</Units>
       <Count>1</Count>
     </item>
+    <item>
+      <EFDB_Name>groupID</EFDB_Name>
```
tribeiro, line 39: Please, rename this `groupId`.

tribeiro (changes requested):
A couple of additional follow up requests.  

I think the Electrometer `stopScan` command does not need the groupId, but the `startScan` command does need it. 

Also, you can now remove the `value` attributes that were previously required. These is no longer needed so you can remove these attributes to cleanup the CSC interface. If you don't want to do this now, please, issue a new ticket to do it later.

The attribute name should be `groupId` and not `groupID`.

tribeiro (approved):
Looks good! thanks for the updates.

PR #876, opened 2024-07-09
MTDome: Add event and enum for sub-system error states.

wvreeven:


wvreeven:
Tiago and I discussed another approach so this PR no longer is needed.

PR #877, opened 2024-07-09
Add ErrorCode enum to LinearStage

couger01:


python/lsst/ts/xml/data/sal_interfaces/LinearStage/LinearStage_Events.xml
```diff
@@ -2,6 +2,17 @@
 <?xml-stylesheet type="text/xsl" href="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsl"?>
 <SALEventSet xmlns:xsi="http://lsst-sal.tuc.noao.edu/schema/SALEventSet.xsd">
   <Enumeration>DetailedState_NotMovingState, DetailedState_MovingState</Enumeration>
+  <Enumeration>
+    ErrorCode_ConnectionFailed,
```
wvreeven, line 6: I see that the python enum explicitly defines the values of the enum. Please do that here as well like so:

```ErrorCode_ConnectionFailed=1,```

and also for the rest of the values of this enum.

wvreeven (approved):
LGTM. Please see my comment before merging.

PR #890, opened 2024-08-14
updated grating enumeration

parfa30:
This should be merged as part of xml-23. Also, when merged, will need to change some of the scripts so that they align with the updated configuration.

### DIFF
--- a/doc/news/DM-45475.misc.rst
+++ b/doc/news/DM-45475.misc.rst
@@ -1,1 +1,0 @@
-Updated grating enumeration for ATMonochromator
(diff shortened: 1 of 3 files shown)


doc/news/DM-45475.misc.rst
couger01, line 1: I would remove the part about Windows 10 and simply leave it as updating the grating enumeration.

couger01 (changes requested):
One requested change.

couger01 (approved):
LGTM.

couger01 (changes requested):
Change the news fragment to use Electrometer section not misc.

parfa30:
I don't know what you mean. Do you mean change the name of the news fragment to be DM-45475.Electrometer.rst? This is for the monochromator.

couger01 (approved):
LGTM.

PR #891, opened 2024-08-14
changed state for ON/OFF for LEDProjector

parfa30:
The states were swapped. Currently, the OFF commands turns the LEDs on.


doc/news/DM-45766.bugfix.rst
couger01, line 1: "Swapped the on/off values"

couger01 (changes requested):
One change.

couger01 (approved):
LGTM.

PR #893, opened 2024-08-20
DM-45874: Update MTMount telemetry with latest version from Tekniker.

tribeiro:


python/lsst/ts/xml/data/sal_interfaces/MTMount/MTMount_Telemetry.xml
```diff
@@ -9,7 +9,6 @@
   <SALTelemetry>
     <Subsystem>MTMount</Subsystem>
     <EFDB_Topic>MTMount_azimuth</EFDB_Topic>
-    <Description>Azimuth axis detail</Description>
```
wvreeven, line 12: Why did you remove this and all other topic descriptions? There is a unit test that checks if all topics have descriptions so that should fail now. See https://github.com/lsst-ts/ts_xml/blob/develop/tests/test_TopicDescription.py#L20-L21 why it doesn't.

wvreeven (changes requested):
Please do not remove the telemetry topic descriptions. There is a unit test that checks for those which now gets skipped for MTMount but that should be fixed at some point.

python/lsst/ts/xml/data/sal_interfaces/MTMount/MTMount_Commands.xml
```diff
@@ -10,6 +10,13 @@
     <Subsystem>MTMount</Subsystem>
     <EFDB_Topic>MTMount_command_openMirrorCovers</EFDB_Topic>
     <Description>Open the mirror covers to allow light in. Warning: This terminology may be the opposite of that used by the Engineering User Interface.</Description>
+    <item>
+      <EFDB_Name>leaf</EFDB_Name>
+      <Description>Specify a leaf to open. By default (leaf=0) open all.</Description>
```
wvreeven, line 15: What other values are possible? I assume it is a single value for a specific leaf. What are the valid values and to which leafs do they correspond? Please introduce an enum for all values and to which leaf they correspond and indicate here that that enum should be used.

wvreeven (approved):
LGTM

PR #894, opened 2024-08-20
Electrometer XML: Mode long to string

parfa30:
Want to command the mode of the electrometer with a string rather than an integer.


couger01 (changes requested):
I don't think this is a good change as having an Int enumeration for the operators and scripts would be easier.

couger01 (changes requested):
Please change the fragment to use the Electrometer section.

couger01 (approved):
LGTM.

PR #901, opened 2024-08-23
Readd UnitToRead enum to Electrometer

couger01:


wvreeven (changes requested):
Please add the missing values to the `UnitToRead` enum and revert the change to the `setMode` command as well.

wvreeven (approved):
LGTM

PR #903, opened 2024-08-28
DM-44022: fix misleading MTM1M3 DetailedState comment

pkubanek:
develop


bstalder (approved):
updated clarification of detailed states approved.

PR #904, opened 2024-09-06
DM-46165: adding optical configuration to TunableLaser XML

parfa30:
Need to move the Optical Configuration enum from ts_TunableLaser to ts_xml so that we can change the optical configuration in scripts. This is necessary because we will swap between F1, F2 and direct for different tests.

### DIFF
--- a/doc/news/interface_changes/DM-46165.tunablelaser.rst
+++ b/doc/news/interface_changes/DM-46165.tunablelaser.rst
@@ -1,0 +1,1 @@
+Adding in Optical Configuration enum for TunableLaser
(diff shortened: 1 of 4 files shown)


tribeiro (approved):
Looks good. 

However, I would suggest you track down the commands/events that this enumeration is used and mention in the attribute description that the field is an enum, e.g;

https://github.com/lsst-ts/ts_xml/blob/a16574f2a257c25557144337a94d9ba50605b017/python/lsst/ts/xml/data/sal_interfaces/TunableLaser/TunableLaser_Events.xml#L45-L46

```
      <EFDB_Name>configuration</EFDB_Name>
      <Description>The new optical alignment configuration; a OpticalConfiguration enum.</Description>
```

tribeiro (changes requested):
Please undo the changes to the lint workflow

tribeiro (approved):

PR #926, opened 2025-01-15
Add the hexapod events.

teweitsai:
Add the **MTHexapod_logevent_highCurrent** and **MTHexapod_logevent_transitionToIdle**.

### DIFF
--- a/python/lsst/ts/xml/data/sal_interfaces/MTHexapod/MTHexapod_Events.xml
+++ b/python/lsst/ts/xml/data/sal_interfaces/MTHexapod/MTHexapod_Events.xml
@@ -411,4 +411,28 @@
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>MTHexapod</Subsystem>
+    <EFDB_Topic>MTHexapod_logevent_highCurrent</EFDB_Topic>
+    <Description>Indicates whether or not that the actuators are experiencing the high current that might come from the unbalanced loading, the cooling greasing, or others when the hexapod stills.</Description>
+    <item>
+      <EFDB_Name>highCurrent</EFDB_Name>
+      <Description>True if the actuators experience the high current. Otherwise, False.</Description>
+      <IDL_Type>boolean</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALEvent>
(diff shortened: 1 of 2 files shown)


python/lsst/ts/xml/data/sal_interfaces/MTHexapod/MTHexapod_Events.xml
```diff
@@ -411,4 +411,28 @@
...
+      <EFDB_Name>highCurrent</EFDB_Name>
+      <Description>True if the actuators experience the high current. Otherwise, False.</Description>
+      <IDL_Type>boolean</IDL_Type>
+      <Units>unitless</Units>
+      <Count>1</Count>
+    </item>
+  </SALEvent>
+  <SALEvent>
+    <Subsystem>MTHexapod</Subsystem>
+    <EFDB_Topic>MTHexapod_logevent_transitionToIdle</EFDB_Topic>
```
tribeiro, line 428: Isn't this information already available in the `MTHexapod_logevent_controllerState` event? Is there really a need to publish just this boolean flag?

Don't want to be picky but I think something like `driversIdle` for a topic name and `idle` for attribute would probably be better.

python/lsst/ts/xml/data/sal_interfaces/MTHexapod/MTHexapod_Events.xml
```diff
@@ -411,4 +411,28 @@
       <Count>1</Count>
     </item>
   </SALEvent>
+  <SALEvent>
+    <Subsystem>MTHexapod</Subsystem>
+    <EFDB_Topic>MTHexapod_logevent_highCurrent</EFDB_Topic>
```
tribeiro, line 416: Are you sure you want to add this to the CSC? why not let the Watcher take care of this?

tribeiro (approved):
Changes looks good, but I think they are a but redundant.

I think there's really no need for the CSC to publish the `highCurrect` event, as this could be done by the Watcher and also, my impression is that the driver state is already covered by another event.

In any case, I am approving the PR so, feel free to ignore my comments above.

teweitsai:
The reason to have these two events is to make the implementation of watcher alarms to have very simple logic. But we can remove these two events and put the complexities to the watcher alarms instead as well. Therefore, I will close this PR. Thanks!

PR #927, opened 2025-01-22
Add descriptions to all atcamera/mtcamera/cccamera items

tony-johnson:
Descriptions have been added to all camera SAL topics.
The exceptions for the camera in  test_TopicDescription.py have been removed.
Many of these descriptions are auto-generated, so the content is what it is.
A few events have been marked with ??? indicating that I am not sure what they are for. I think these events are probably never issued and I will look into removing them (under a separate JIRA)


rbovill (approved):
LGtM

PR #930, opened 2025-01-29
Filter exchange system changes for Cycle 40

aboucaud:
PR addressing changes required by the filter exchange system before Cycle 40 and described at
https://rubinobs.atlassian.net/browse/CAP-1064
including the addition of the new `g0` unit definition
https://rubinobs.atlassian.net/browse/CAP-1067


rbovill (approved):
LGtM

PR #933, opened 2025-02-04
Update Filter exchange system telemetry

aboucaud:
This is a direct consequence of fixing the fcs data dictionary to remove the registration of fake sensors that never published, which went unseen up to now.

This file was created with 
org-lsst-ccs-camera-sal-xml for the cycle40 SAL release.


rbovill (approved):
LGtM

PR #934, opened 2025-02-06
DM-48784 Revisions for DIMM CSC

bbrondel:
The DIMM CSC has some shortcomings and some discrepancies between the XML and the actual implementation of the DIMM.


rbovill (approved):
LGtM

PR #935, opened 2025-02-10
Implement CAP-1062 (Generic command)

tony-johnson:
This implements a generic command to allow us to issue "arbitrary" commands to the camera
Commands will only be accepted if the camera configuration allows it.
This is intended to give us a way to allow commands if an urgent need is discovered during commisioning/operations without waiting for an XML release.


rbovill (approved):
LGtM

PR #936, opened 2025-02-11
Commit MTCamera_Events.xml with events previously missed from CAP-1064

tony-johnson:
Updates to MTEvents.xml for filter changer, previously missed from CAP-1064


rbovill (approved):
LGtM

PR #947, opened 2025-03-12
DM-49413 Add ringssMeasurement event to ESS

bbrondel:
This event makes it possible to load SOAR RINGSS data into the EFD.


tribeiro (approved):
Looks good..

PR #974, opened 2025-06-02
DM-51223 Update DREAM interface so that it is indexed.

bbrondel:
Backs out PR #959 and makes DREAM an indexed CSC, consistent with past deployments.


tribeiro (approved):
This PR is reverting a change that was inadvertently made to the DREAM interface. We originally made it indexed in case we ever have more than one of the DREAM system and we would like to retain this capability, even if we currently do not have any plans for more than one instance of it.

PR #975, opened 2025-06-13
DM-51385: Fix M1M3 thermal scanner 3 position values

elanaku:


pkubanek (approved):
Looks good. Thanks for spotting that.

PR #976, opened 2025-07-16
Add preclipped limit configuration.

pkubanek:


bbrondel (approved):
LGTM aside from the strange CI error and the need for a news fragment.

PR #981, opened 2025-08-01
Add visitId to MTAOS events

gmegh:


tribeiro (approved):
Looks good, but you might want to update the news fragment to reflect all the changes you've made to the MTAOS xml. You only included the visit id but you are adding more things now.

PR #1003, opened 2025-09-15
OSW-670: Add command to flush the Scheduler queue and additional events with information on the Scheduler internal queue.

tribeiro:


wvreeven (approved):
LGTM

PR #1006, opened 2025-09-16
Implement CAP-1085

tony-johnson:
change enumerations to be compatible with watcher implementation
add Java as a runtime language for watcher


wvreeven (approved):
Looks good to me.

PR #1013, opened 2025-09-26
OSW-1141: Add test that rejects SAL timestamp attributes with type float.

bbrondel:
The added test finds timestamp attributes by two criteria: the name contains the substring "timestamp" (case insensitive) or the description contains the word "TAI", also case insensitive.


rbovill (approved):
Looks great. Thanks for creating this unit test!

PR #1078, opened 2026-05-05
OSW-2266: Rollback ESS_agcGenset150 fields extension in favor of new telemetry ESS_agcMains150.

sebastian-aranda:


wvreeven (approved):
Looks good but I think `kWh` is preferred over `kW * h`.

PR #1084, opened 2026-05-15
OSW-1555: Add MTDome tables.

bbrondel:
See discussion in lsst-ts/ts_eas#83.

### DIFF
--- a/python/lsst/ts/xml/tables/mtdome/louver_table.py
+++ b/python/lsst/ts/xml/tables/mtdome/louver_table.py
@@ -1,0 +1,127 @@
+# This file is part of ts_xml.
+#
+# Developed for Vera Rubin Observatory.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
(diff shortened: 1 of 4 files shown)


python/lsst/ts/xml/tables/mtdome/louver_table.py
```diff
@@ -0,0 +1,114 @@
...
+    ----------
+    index : int
+        Position of the louver in `LouverTable`, 0 - 33.
+    name : str
+        Louver name, e.g. "A1", "B2", "N2".
+    azimuth : float
+        Angular offset (degrees) of the louver normal relative to the dome
+        azimuth.
+    type_number : int
+        Louver type identifier, 1 - 8. See EIE-DWG-45-10-00-00.
```
pkubanek: Can that be enum? Do we have additional data explaining meaning of the codes? If yes, please make it enum - otherwise leave it as it is now.

pkubanek (changes requested):
Please add tests - see test_tables_m1m3.py for how M1M3 tables tests are handled. A test showing find_louver works (both positive, when name exists, as well as negative, when searched name doesn't exists) shall be enough.

pkubanek (approved):
Thanks

PR #1085, opened 2026-05-29
Changed order of enumeration elements for field currentSeverity

mxturri:


wvreeven (changes requested):
Please make sure that the enumeration values match the AlarmSeverity values, so use NONE instead of NOMINAL. I would prefer to uncomment and update the highestSeverity values as well.

wvreeven (approved):
LGTM

PR #1086, opened 2026-06-02
Changed type of currentSeverity and highestSeverity fields for alertR…

mxturri:
…aised event back to long (was int).


wvreeven (approved):
LGTM

PR #1088, opened 2026-06-12
tickets/OSW 2379

couger01:
- **Add LEDProjector ledsOn event**
- **fixup! Add LEDProjector ledsOn event**


teweitsai (approved):
Looks good.

PR #1091, opened 2026-06-22
Add dimensionless rain level and light level topics for ESS.

bbrondel:
This change adds `rainLevel` and `lightLevel` topics as per a discussion with Wouter regarding the Aurora Cloud Sensor, which has telemetry on these topics. The `rainRate` topic already exists, but it specifies units whereas the Aurora Cloud Sensor does not document units associated with its data. The new topics are modeled after `rainRate`, but they omit `sensorName` and `location`.


wvreeven (changes requested):
All ESS topics require sensorName and location so please add those. Note that not all ESS topics currently have them and we're working on adding them.

wvreeven (approved):
LGTM

PR #1092, opened 2026-06-25
Update the OperatingMode enum.

wvreeven:


sebastian-aranda (approved):
I'm a bit hesitant about e7a89b79f9eb3e5168fa3893d02fe2aab2a65c1e, but the rest looks good, thanks for these changes 🙌 ! I we are keeping e7a89b79f9eb3e5168fa3893d02fe2aab2a65c1e: it would be good to reference the W3000 manual docushare link as a comment to the OSW-2550 ticket for easy tracking.

